PR #25, opened 2016-10-18
Use redirect method for mobile at login from Github

jackypig:
screen shot 2016-10-18 at 11 42 52 am
screen shot 2016-10-18 at 11 43 03 am


src/Utils.test.tsx
```diff
@@ -0,0 +1,14 @@
+import utils from "./Utils";
```
michaelmyers: Can you order the imports as outlined [here](https://github.com/bespoken/dashboard/blob/master/docs/patterns.md#imports) for consistency? 

Some other files still need to be changed but this is recommended for new files.

PR #47, opened 2016-10-31
Login redirect

### LINKED ISSUE #39: Redirect to Requested Resource After Login
If a user attempts to access a resource that requires authentication with a direct link and are not logged in, they will be redirected to the login screen.  Once logged in, they are then redirected to the dashboard index.  We would like them to be redirected to the original requested link after the login instead of the dashboard index.

Current:

`/dashboard/some/protected/resource` --> `/dashboard/login` --> `/dashboard`

Desired:

`/dashboard/some/protected/resource` --> `/dashboard/login` --> `/dashboard/some/protected/resource`

chrsdietz:
Fixes #39 

- The "session.ts" uses a strategy method to figure out where to go next after login.  Passing an "undefined" is going to use the default strategy which goes to the base dashboard like it was originally.
- The "Login.tsx" now accepts a redirect url to move either go back or redirect to another page.  Passing "Back" will call back on the browser to go to the page it was before requiring login.

### DIFF
--- a/src/pages/LoginPage.tsx
+++ b/src/pages/LoginPage.tsx
@@ -3,7 +3,7 @@ import * as React from "react";
 import { connect } from "react-redux";
 
 import { changeForm } from "../actions/authForm";
-import { login, loginWithGithub } from "../actions/session";
+import { login, loginWithGithub, ToPathStrategy } from "../actions/session";
 import AuthForm from "../components/AuthForm";
 import Card from "../components/Card";
 import { Cell, Grid } from "../components/Grid";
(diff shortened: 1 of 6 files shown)


src/pages/LoginPage.tsx
```diff
@@ -13,9 +13,14 @@ interface LoginPageProps {
     email: string;
     password: string;
     error: string;
+    /**
+     * The redirect URL to send to after successful login. You can also input "back" to indicate that the
+     * login page should return to the page that sent it.
+     */
+    redirect?: string;
```
chrsdietz: Wasn't sure if this was the best way to go about it, but basically you can put a relative path like "/logs" and it'll send you to the logs page.  Or you just put "back" and it'll go back to where you were.

PR #50, opened 2016-11-02
Skill Service, Skill List, Create Skill

### LINKED ISSUE #28: Once logged out the login page has pre-filled username and password which overlaps with "email" and "password"
Once the user logs out of the dashboard, the login page is pre-filled with the last used username and password and it is overlapped with the field word "email" and "password"

image

### LINKED ISSUE #38: HTTPS not HTTP
We need the logging server to be HTTPS, without it we will get Mixed Content errors when we push the dashboard to production.

### LINKED ISSUE #42: Migrate Remaining Tests to Enzyme
There are some remaining tests that still use "react-addons-test-utils", Content.test.tsx for instance.  We want to remove this dependency and instead use "enzyme" (see Navigation.test.tsx as an example).

### LINKED ISSUE #46: Simple Skills Service
Simple service to provision new skills.

Most likely will want to call this a lambda service underneath the covers.

michaelmyers:
Fixes #38 
Fixes #42
Fixes #28 
Fixes #46 

- Fixes a couple of bugs
- Adds a generic Source (the generic form of Skill) Service
- Adds ability to create a new skill
- Adds a list to view your skills
- Updates log page to view your skill

@chris @umut-genlik 

Please note, I am using the word "Skill" for public facing terminology in the app however underneath it is using the Source object that has a Alexa Profile.  We will want to expand on the profiles later but they will allow us to apply different UI logic for displaying the data based on what profile the source has.


src/index.tsx
```diff
@@ -36,18 +40,27 @@ let firebaseConfig = {
...
 Firebase.initializeApp(firebaseConfig);
-Firebase.auth().onAuthStateChanged(function(user: Firebase.User) {
-    console.log("onAuthStateChanged");
-    console.log(user);
+Firebase.auth().onAuthStateChanged(function (user: Firebase.User) {
+    let firebaseInitializeTime = +new Date() - +firebaseInitializeTimer;
+    console.log("Firebase took " + firebaseInitializeTime + "ms to inialize");
+    // If there is a user, set it
+    if (user) {
+        store.dispatch(setUser(new FirebaseUser(user)));
```
michaelmyers: There is a race condition here with Firebase where it doesn't return the User immediately after you initialize, see my notes [here](https://github.com/bespoken/dashboard/blob/master/docs/issues.md#firebase--currentuser-race-condition).

When the page renders, we start loading resources for the user from Firebase and if we call render before we have a user, we will not have proper authentication from Firebase to start making requests to private resources.  

So, we have to wait for Firebase to tell us there is a user, then we set the user and render the dashboard.  

We were also able to remove [this](https://github.com/bespoken/dashboard/blob/master/src/reducers/session.ts#L16) dependency within the session reducer.

src/index.tsx
```diff
@@ -36,18 +40,27 @@ let firebaseConfig = {
...
+    }
+    // We need to wait for the user to be available before we can render the app
+    render();
```
michaelmyers: We wrap the setUser in a null check so we don't set it if one isn't passed back.  The initial session state has the user as not defined so when we render it and they are trying to access the dashboard, [this](https://github.com/bespoken/dashboard/blob/master/src/index.tsx#L52) code checks to see if there is a user and if there isn't we dump them to the login.

src/index.tsx
chrsdietz: If there's a callback for then when the user is ready, it may be more elegant to modify [auth.login](https://github.com/bespoken/dashboard/blob/skill-service/src/services/auth.ts#L50) so that the callback is not returned until everything is verified to be ready. I just feel that we have this login mechanism abstracted away from the UI logic, but here we're setting the user on the UI page. Technically we could log anywhere and would always have to set the user.

src/pages/LogsPage.tsx
```diff
@@ -73,25 +79,61 @@ export class LogsPage extends React.Component<LogsPageProps, any> {
...
+                this.props.getLogs(source.secretKey);
+                this.setState({
+                    source: source
+                });
+            }
+        }
+    }
+
+    componentWillReceiveProps(nextProps: LogsPageProps, nextContext: any) {
+        if (this.state.source === undefined) {
```
michaelmyers: As discussed, there is a bigger issue here around how we request logs and if we try to keep them.  I have made an [issue](https://github.com/bespoken/dashboard/issues/51) for this.

PR #61, opened 2016-11-08
Updates to Log Page

### LINKED ISSUE #33: Logging Request View and Response View
Two views, one for requests and one for responses.

### LINKED ISSUE #59: Log Conversation View
Parse the logs into conversations.

michaelmyers:
Fixes #59, Fixes #33

- Move skill title to header
- Styling to the logs page
- New utils for numbers and colors, used for programmatically setting the color of the user avatar

@umut-genlik @chrsdietz @jackypig


michaelmyers:
Ready to go!

screen shot 2016-11-08 at 11 22 45 pm

src/components/ConversationList/ConversationList.tsx
```diff
@@ -0,0 +1,112 @@
...
+                }
+
+                if (log.tags.indexOf("response") > -1) {
+                    response = log;
+                }
+
+                if (response && request) {
+                    if (response.transaction_id === request.transaction_id) {
+                        conversations.push(new Conversation({ request: request, response: response }));
+                        request = undefined;
```
chrsdietz: The request and response will only get reset if `response.transactin_id === request.transaction_id`.  Then in the next loop, you could possibly get a `request` or `response` from the previous log if they don't match the criteria above.

src/components/ConversationList/ConversationList.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    render() {
+
+        let conversations: JSX.Element[] = [];
+
+        for (let conversation of this.state.conversations) {
+            conversations.push((
+                <ConversationListItem
+                    key={conversation.id}
```
chrsdietz: We don't need to for this pull request, but we may need some kind of Adapter mechanism that recycles `ConversationListItems` as the user scrolls through in case there are thousands of elements.

src/components/ConversationList/ConversationListItem.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+describe("ConversationListItem", function () {
+    it("renders correctly", function () {
+        let onClick = sinon.spy();
+        let conversation = new Conversation({ request: requestIntentLog, response: responseLog });
+        let wrapper = shallow(<ConversationListItem conversation={conversation} onClick={onClick} />);
+
+        expect(wrapper.find("li")).to.have.length(1);
+        let li = wrapper.find("li");
+        expect(li.props().style.backgroundColor).to.equal("#E0E0E0");
```
chrsdietz: Do we care that the background is grey?  This seems volatile. I thought these "renders correctly" tests were to just ensure it doesn't crash when starting.

src/pages/LogsPage.tsx
```diff
@@ -82,17 +90,24 @@ export class LogsPage extends React.Component<LogsPageProps, LogsPageState> {
...
+        // then we find the one with the matching slug
         for (let source of sources) {
             if (this.props.params.sourceSlug === source.slug) {
+                this.props.setCurrentSource(source);
                 this.props.getLogs(source.secretKey);
                 this.setState({
-                    source: source
+                    source: source,
+                    request: this.state.request,
+                    response: this.state.response
```
chrsdietz: Are we setting the state potentially dozens of times or can we break out of the loop from this point?

src/components/ConversationList/ConversationListItem.test.tsx
michaelmyers: This is checking the default behavior of a list item when the list item is not active.  The background color changes if active=true, which is tested below.

src/pages/LogsPage.tsx
michaelmyers: Added a `break` to jump out when we find a match.

src/components/ConversationList/ConversationList.tsx
michaelmyers: Yes, this logic is a bit fragile.  Created ticket #65

PR #63, opened 2016-11-09
Error checks

chrsdietz:
Addresses some of #63 but not completely closed yet.

- Adding string validator to skill name
- Names can no longer be shorter than three characters and can not contain special characters except hyphen
- Adding the pattern to the input field


src/pages/NewSourcePage.tsx
```diff
@@ -17,6 +17,16 @@ interface NewSourceProps {
     sources: Source[];
 }
 
+/**
+ * Validator class for the SourceForm.  Exported for direct testing.
+ */
+export class OurNameRule implements NameRule {
```
michaelmyers: Maybe us a name that is a little more descriptive such as `SourceNameRule`

PR #87, opened 2016-11-14
Logs logic

### LINKED ISSUE #51: Logs Page and When to Request New Logs
The LogsPage has some weird logic about when it requests logs, there is probably a better way of doing this.

chrsdietz:
Fixes #51 

- `index.tsx` now will pre-load the sources and convos when entering all `/skill/:sourceId/` paths.
- The `current selected source` is now set before entering any `/skill/:sourceId/` route and unset every time the user leaves.
- Separated out the conversation panel in to a new component called `Interaction`. 
- `Logs` now only cares about a single Source instead of the entire list.


src/pages/LogsPage.tsx
```diff
@@ -220,12 +104,9 @@ export class LogsPage extends React.Component<LogsPageProps, LogsPageState> {
                     </div>
                 </Cell>
                 <Cell col={6} style={{ maxHeight: this.getContentHeight() - 30, overflowY: "scroll" }}>
-                    <h6>REQUEST</h6>
-                    {this.getRequest()}
-                    <h6>CONSOLE</h6>
```
michaelmyers: The `<OutputList />` needs to also be included in `<Interaction />`

PR #101, opened 2016-11-23
Infinite scroll

### LINKED ISSUE #64: Recycle the ConversationListItem Views
Instead of creating new ConversationListItems in the ConversationList for all the Conversation objects, we should instead use a recycler to just update the views that already exist.

chrsdietz:
Fixes #64 

- Added an infinite list scroller for recycling elements.
- Fixed padding and clipping on logs page.
- Unregistering listeners on unmounting.

### DIFF
--- a/src/components/ConversationListView/ConversationListView.tsx
+++ b/src/components/ConversationListView/ConversationListView.tsx
@@ -28,11 +28,6 @@ export default class ConversationListView extends React.Component<ConversationLi
         };
     }
 
-    divStyle(): React.CSSProperties {
-        return {
-        };
-    }
-
     listStyle(): React.CSSProperties {
         return {
             marginLeft: "20px",
--- a/src/components/ConversationListView/ConversationListView.tsx
+++ b/src/components/ConversationListView/ConversationListView.tsx
@@ -6,6 +6,7 @@ import Conversation from "../../models/conversation";
 import ConversationList, { ConversationMap } from "../../models/conversation-list";
 import ConversationListViewItem from "./ConversationListViewItem";
 
+// TODO: Would be nice to get this to work with typescript
 let ReactList = require("react-list");
 
 export interface ConversationListViewProps {
@@ -73,7 +74,6 @@ export default class ConversationListView extends React.Component<ConversationLi
     }
 
     renderItem(index: number, key: string): JSX.Element {
-        console.info("Render item " + index + " " + key);
         let conversation = this.props.conversations[index];
         return <ConversationListViewItem
             key={conversation.id}
@@ -84,8 +84,6 @@ export default class ConversationListView extends React.Component<ConversationLi
     }
 
     render() {
-        console.info("Found " + this.props.conversations.length);
-
         return (
             <div>
                 {this.props.conversations.length > 0 ? (
--- a/src/components/ConversationListView/ConversationListView.tsx
+++ b/src/components/ConversationListView/ConversationListView.tsx
@@ -95,7 +95,8 @@ export default class ConversationListView extends React.Component<ConversationLi
                         itemRenderer={this.renderItem.bind(this)}
                         length={this.props.conversations.length}
                         pageSize={this.props.conversations.length} // TODO: paging needs to be fixed so it doesn't load all elements at once.
-                        type={"uniform"} />
+                        type={"uniform"}
+                        useStaticSize={true} />
                 ) : (
                         <p> No available data </p>
                     )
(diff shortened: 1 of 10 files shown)


src/components/ConversationListView/ConversationListView.tsx
```diff
@@ -24,10 +28,15 @@ export default class ConversationListView extends React.Component<ConversationLi
         };
     }
 
-    style(): React.CSSProperties {
+    divStyle(): React.CSSProperties {
```
michaelmyers: This isn't used.

src/pages/LogsPage.tsx
```diff
@@ -44,40 +46,58 @@ function mapDispatchToProps(dispatch: Redux.Dispatch<any>) {
 
 export class LogsPage extends React.Component<LogsPageProps, LogsPageState> {
 
+    resizeEvent: browser.WrappedEvent;
+
     constructor(props: LogsPageProps) {
         super(props);
-
         this.state = {
+            width: 0,
```
michaelmyers: Width and height should be queried directly and not held in the state.  The data already exists on the element itself and should be queried when needed, something else may modify it's size and the data in the state will then be out of sync with the actual height and width.

src/pages/LogsPage.tsx
```diff
@@ -44,40 +46,58 @@ function mapDispatchToProps(dispatch: Redux.Dispatch<any>) {
...
+    }
 
-            // The case where the browser got bigger
-            if (target.innerWidth >= browser.mobileWidthThreshold && this.previousSize.width < browser.mobileWidthThreshold) {
-                this.forceUpdate();
-            }
+    updateDimensions() {
+        // Algorithm taken from https://andylangton.co.uk/blog/development/get-viewportwindow-size-width-and-height-javascript
+        // Modified to get around unit tests which don't have half this.
+        let width: number, height: number;
```
michaelmyers: This algorithm is a better version of what is available by  `browser. size()`.   I would recommend enhancing that one and removing the dependency of `window` from LogsPage element.

You will need to change the class back to element `body`

src/pages/LogsPage.tsx
```diff
@@ -44,40 +46,58 @@ function mapDispatchToProps(dispatch: Redux.Dispatch<any>) {
...
+            width = 200;
+            height = 200;
+        }
 
-            // Update the previous size
-            this.previousSize = {
-                width: target.innerWidth,
-                height: target.innerHeight
-            };
+        this.setState({
```
michaelmyers: Setting the state causes a re-render of the component, I would avoid updating it this frequently.

src/pages/LogsPage.tsx
```diff
@@ -95,38 +117,28 @@ export class LogsPage extends React.Component<LogsPageProps, LogsPageState> {
...
+            width: this.state.width,
+            height: this.state.height,
             source: this.state.source,
             request: conversation.request,
             response: conversation.response,
             outputs: conversation.outputs
         });
     }
 
-    getContentHeight() {
```
michaelmyers: I would move a method like this to the `browser` util and use it to get the height when needed

src/pages/LogsPage.tsx
chrsdietz: This has to be cached in state because these elements are not available at render time (until rendering is complete at least) so it couldn't be retrieved in `render()`. As it currently works, this is updated every time the browser resizes and can only be changed by that.

src/components/ConversationListView/ConversationListView.tsx
chrsdietz: Removed

src/pages/LogsPage.tsx
chrsdietz: "browser.size()" is described as "getting the size of the browser window" while this algorithm retrieves the size of the DOM element if it exists.

chrsdietz: New method now only looks at the root element rather than the entire DOM tree.  Should be faster and more reliable now, but now the utility method may not be needing update for this.

src/pages/LogsPage.tsx
chrsdietz: This method now turned in to `getDimentions()` method.  It looks at the size of the root element and only uses Window as a fallback.  Need re-review.

src/pages/LogsPage.tsx
chrsdietz: Ok.  Upon further testing.  The only thing that mattered was the window dimensions anyway so all of what I just said is completely worthless.  Now only window dimensions are being read and the max window size is being adjusted based on the offset from the corner.

PR #103, opened 2016-11-29
Auth promise

chrsdietz:
- Converting the low-level `auth` class to use Promises instead of callbacks.


src/actions/session.test.ts
```diff
@@ -13,6 +13,12 @@ const mockStore = configureMockStore(middlewares);
 
 let expect = chai.expect;
 
+let user = new User({
+    email: "test@test.com",
+    displayName: "TestyMcTestFace",
```
michaelmyers: 👍

PR #107, opened 2016-12-06
Filter

### LINKED ISSUE #73: Ability to Filter Current Logs
After a query, a user might want to filter the existing queried logs by additional parameters without having to perform new queries.

The filter can be potentially on any piece of metadata such as userId or intent type.  The filter can be applied by simply selecting a piece of information.

chrsdietz:
Fixes #73 

-  Adding the "select" component which is a widget that acts like a menu, but is a text field
- Adding the "react-toolbox" which will eventually replace the other components that we have.
-  Adding the "date picker" component which is the ability select dates.
-  Adding the ability to filter logs by their conversation ID (will most likely be removed).
-  Adding ability to filter logs by their log type.
-  Adding ability to filter logs by their day.


harness/index.html
```diff
@@ -1,51 +1,58 @@
...
+  <link rel="stylesheet" href="/style.css">
+  <script src="/style/script/getmdl-select.js"></script>
+
+</head>
+<body>
+  <!-- This is where the dashboard will be injected -->
+  <div id="dashboard"></div>
+
+  <!-- Dashboard Required Dependencies -->
+  <script src="https://unpkg.com/react@15.3.2/dist/react.js"></script>
```
michaelmyers: Since these are brought in as dependencies within webpack, are they still needed here?

chrsdietz: Those were already there I think.  They just shifted around.

src/pages/logspage/LogsPage.tsx
```diff
@@ -156,15 +148,15 @@ export class LogsPage extends React.Component<LogsPageProps, LogsPageState> {
...
         return (
             <div
-                ref={ this.onRootLayout.bind(this) }>
+                ref={this.onRootLayout.bind(this)}>
                 <Grid
                     noSpacing={true}>
                     <Cell col={6} phone={4} tablet={4} style={{ paddingLeft: "10px", paddingRight: "5px" }}>
-                        <div style={{ maxHeight: this.state.lastDimens.cellDimens.height, overflowY: "scroll" }}>
-                            <ConversationListView
+                        <div>
```
michaelmyers: I do not think you need this `<div>` wrapping the `<FilterableConversationList>`

src/utils/promise-utils.tsx
```diff
@@ -0,0 +1,25 @@
+/**
+ * Promise that will filter out items in a list.  If the filter is undefined, then the original list is returned.
+ * If there are no items found, then the promise will be rejected.
+ */
+export function filter<T>(items: T[], filter: (item: T, index: number) => boolean): Promise<T[]> {
```
michaelmyers: The file name `promise-utils` is inconsistent with the rest of the utils.  I only care about consistency, you can change them all to `-util` if you like.  If that is what you want to do that should probably be a separate ticket.

webpack.config.js
```diff
@@ -42,13 +54,4 @@ module.exports = {
     new ExtractTextPlugin("style.css", { allChunks: true })// ,
     // Not ready for this yet // new webpack.optimize.UglifyJsPlugin()
   ],
-
-  // When importing a module whose path matches one of the following, just
-  // assume a corresponding global variable exists and use that instead.
-  // This is important because it allows us to avoid bundling all of our
-  // dependencies, which allows browsers to cache those libraries between builds.
-  externals: {
```
michaelmyers: This is where I am a bit confused.  When you remove these as [externals](https://webpack.js.org/configuration/externals/#externals), webpack will then bundle these dependencies in with the rest of the library.  However, when I tried to remove them from the harness index.html (line 52), I get a cannot find React error.  Do you understand what is going on here?

src/pages/logspage/FilterableConversationList.tsx
```diff
@@ -0,0 +1,109 @@
+import * as React from "react";
```
michaelmyers: I understand keeping this component and LogsFilterComponent close to LogsPage however they seem like they can be reusable in other places.  Should they instead live under `src/components`?  Any reason you put them here?

src/pages/logspage/Filters.tsx
```diff
@@ -0,0 +1,79 @@
+import * as moment from "moment";
```
michaelmyers: Similar to my comment about the LogsFilterComponent above, this looks like it should be in `models` so it can be reused in other places.

webpack.config.js
chrsdietz: As far as I can tell, when we had `react` in both the webpack.config and the `package.json` it was pulling two different libraries down at the same time and causing the javascript error.

src/pages/logspage/FilterableConversationList.tsx
chrsdietz: Only that it can only filter `conversations` right now and it was very unique to the Logs page. For that matter, `ConversationList` is also really unique to the Logs page.

src/pages/logspage/Filters.tsx
chrsdietz: As stated earlier.  They only accept `Conversations` right now which is very limiting.  The interface could be used in other locations.

harness/style/script/getmdl-select.js
```diff
@@ -0,0 +1,25 @@
+{
```
chrsdietz: It'll need to be included somewhere, yes.

chrsdietz: Although it may be possible to hook it in to react component somehow, but I'm not sure how to do that.

Issue #110, opened 2016-12-08
Display Stack Field for Crashes


michaelmyers:
We are not reading in the stack field on the log payload which contains crash information.  This should be displayed in the console view.

PR #115, opened 2016-12-15
Display Crashes and Source Information

### LINKED ISSUE #54: Ability to Access the Secret Key After Skill Creation
Currently, there isn't a way to get your secret key again after you created the skill.

michaelmyers:
Fixes #54, #110 

- New StackTrace & StackTraceElement object and ability to display them in the interaction
- New LogQuery object for building queries and keeping track of the current log query
- Removal of Drawer
- Index Route is no longer the preview message, instead it is your list of Skills
- SourcePage for viewing a summary of your source/skill information


michaelmyers:
Not ready for a review or merge yet.

src/components/SourceSummaryView.tsx
```diff
@@ -0,0 +1,86 @@
...
+                            </Cell>
+                            <Cell col={4}>
+                                <FormInput
+                                    type={"text"}
+                                    style={{paddingBottom: 0}}
+                                    value={this.props.sourceSummary.totalCrashes.toString()}
+                                    label={"Crashes"}
+                                    floatingLabel={true}
+                                    autoComplete={"off"}
+                                    readOnly={true} />
```
DeaVea: It may be verbose, but would it be easier to maintain if we split this up in to individual components?  All of these FormInputs are basically the same except for their label.

src/models/conversation-list-summary.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    private conversationList: ConversationList;
+
+    readonly startTime: Date;
+
+    readonly endTime: Date;
+
+    get uniqueUsers(): string[] {
+
+        let usersMap: { [userId: string]: string } = {};
```
DeaVea: Since this is an immutable object, we should be able to cache the unique users so we're not querying it every time this identity is called.

src/models/conversation-list-summary.ts
```diff
@@ -0,0 +1,98 @@
...
+                for (let stackTrace of conversation.stackTraces) {
+                    crashes.push({
+                        timestamp: stackTrace.timestamp,
+                        stackTrace: stackTrace
+                    });
+                }
+            }
+        }
+
+        return crashes;
```
DeaVea: We can cache these too.

src/utils/date.ts
```diff
@@ -0,0 +1,29 @@
+import * as moment from "moment";
+
+export namespace DateUtil {
+    export function daysAgo(days: number): Date {
+        let newDate = new Date();
+        newDate.setDate(newDate.getDate() - days);
+        return newDate;
```
DeaVea: Isn't this the kind of stuff "moment" is used for?

DeaVea: Should we just stop using "Date" all together?

DeaVea:
"Crash" reports don't get filtered in "Error" filter.

DeaVea:
While looking at logs, selecting a different project in the toolbar will take you out of logs and in to a summary. It seems like it should keep you in logs given the context.

DeaVea:
These select inputs shouldn't be focusable if you can't change them. Also that bar, but making them unfocusable should fix that.  

screen shot 2016-12-21 at 10 55 25 am

DeaVea:
The "show" button doesn't change to a pointer when you go over it.

DeaVea:
I don't think this is a bug in the PR, but a general bug.  The "select project" piece in the header doesn't differentiate two projects with the same name.

Reproduce:

Create two projects with the same name.
Swap between the two projects using the header.

Expected result:

The logs should be swapped.  Their ID in the URL should change.  The Secret Key should change.

Actual result:

It always takes the top one and swaps to it.

screen shot 2016-12-21 at 11 10 52 am

DeaVea:
#124 addresses the issue outlined above.

michaelmyers:
I have accounted for all comments except the crash filter, which I had made a ticket for: https://github.com/bespoken/dashboard/issues/125

PR #142, opened 2016-12-29
Goodbye getmdl-select, hello Dropdown

michaelmyers:
Removes the getmdl-select based dropdown menu and replaces it with the react-toolbox Dropdown.

This starts to put us in a place where we can start leveraging SCSS-modules for style and share certain shared styles.


src/frames/Dashboard.tsx
```diff
@@ -6,18 +6,33 @@ import { push, replace } from "react-router-redux";
...
 import UserControl from "../components/UserControl";
 import { CLASSES } from "../constants";
 import Source from "../models/source";
 import User from "../models/user";
 import { State } from "../reducers";
 
+/**
+ * Simple Adapter so a Source can conform to Dropdownable
+ */
+class SourceDropdownableAdapter extends Source implements Dropdownable {
```
chrsdietz: It's probably best to have the SourceDropDownableAdapter to contain Source rather than extending Source.

src/frames/Dashboard.tsx
```diff
@@ -94,27 +98,36 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+  indexOf(source: Source): number {
     if (source) {
-      let adapter = this.state.adapter;
-      let maxCount = adapter.getCount();
-      for (let i = 0; i < maxCount; ++i) {
-        if (adapter.getItem(i).id === source.id) {
-          return i;
+      for (let item of this.props.sources) {
+        if (item.id === source.id) {
+          return this.props.sources.indexOf(item);
```
chrsdietz: I think the way `indexOf` works is it scans through the elements until it finds it, so we're essentially iterating through the array twice. The previous method returned the index with one iteration.

src/components/MDLComponent.tsx
```diff
@@ -25,12 +25,7 @@ declare let window: MDLWindow;
 export default class MDLComponent<P, S> extends React.Component<P, S> {
     componentDidMount() {
         if (window.componentHandler) {
-            window.componentHandler.upgradeDom();
-            // The preferred method is to upgrade the individual elements
-            // However the Select component was not working all the way
-            // For now, we will call upgradeDom, which upgrades the whole dom
-            // and does not seem to impact performance at the moment. -- MMM
-            // window.componentHandler.upgradeElement(findDOMNode(this));
+            window.componentHandler.upgradeElement(findDOMNode(this));
```
chrsdietz: I had ripple effects in a list view that stopped working with this change.  I'm not sure what's the cause of it.  `upgradeDom()` works though.

chrsdietz: I figured out it's because with the `upgradeDom`, all elements get upgraded, but with this method only the top level dom gets upgraded.  I'm updating my branch with a `Ripple` class that makes this easier to add.

PR #149, opened 2016-12-30
FilterBar Style and Mobile Hide

michaelmyers:
Fixes #

- LogsExplorer now contains the logic LogsPage used to
- Styling on FilterBar to reduce the size
- FilterBar autohides on mobile when you scroll


src/components/Measure.tsx
```diff
@@ -11,18 +11,23 @@ export interface MeasureProps {
...
+    private root: Element;
+    private resizeEvent: browser.WrappedEvent;
+    private observer: MutationObserver;
 
     componentDidMount() {
         this.resizeEvent = browser.onResize(this.updateDimensions.bind(this));
         this.resizeEvent.register();
         this.updateDimensions();
+
+        this.observer = browser.onMutation(this.root, this.onMutation.bind(this));
```
chrsdietz: The observer doesn't seem to be disconnected on unmount.

src/utils/browser.test.ts
```diff
@@ -82,6 +82,13 @@ describe("Utils", function () {
         });
     });
 
+    describe("onMutation", function() {
+        it("returns undefined on node", function() {
```
chrsdietz: This test seems to be useless because "MutationObserver" doesn't exist so it's always going to return undefined. Probably should use Sinon to stub it.

src/components/Measure.tsx
michaelmyers: Now disconnects in componentWillUnmount

PR #168, opened 2017-01-07
Fix for when a Request Doesn't Exist on a Conversation

michaelmyers:
Got some data without requests, this checks it so make sure they exist.

### DIFF
--- a/src/models/conversation.ts
+++ b/src/models/conversation.ts
@@ -27,7 +27,7 @@ export default class Conversation implements ConversationProperties {
         this.stackTraces = props.stackTraces ? props.stackTraces.slice() : [];
     }
 
-    get id(): string {
+    get id(): string | undefined {
 
         let id: string;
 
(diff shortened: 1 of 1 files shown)


src/models/conversation.ts
```diff
@@ -28,7 +28,16 @@ export default class Conversation implements ConversationProperties {
     }
 
     get id(): string {
```
chrsdietz: Should it return "string | undefined" to tell the caller that an undefined is possible?

src/models/conversation.ts
```diff
@@ -28,7 +28,16 @@ export default class Conversation implements ConversationProperties {
...
+        if (this.request) {
+            id = this.request.id;
+        } else if (this.response) {
+            id = this.response.id;
+        }
+
+        return id;
     }
 
     get applicationId(): string {
```
chrsdietz: Same as previous comment.

src/models/conversation.ts
```diff
@@ -108,23 +117,32 @@ export default class Conversation implements ConversationProperties {
...
     get intent(): string | undefined {
-        if (this.request.payload.request && this.request.payload.request.intent) {
+        if (this.request && this.request.payload.request && this.request.payload.request.intent) {
             return this.request.payload.request.intent.name;
         } else {
             return undefined;
         }
     }
 
     get timestamp(): Date {
```
chrsdietz: This one should return " undefined" as well.

Issue #170, opened 2017-01-07
When today's date is selected as starting date, then no data displays


pathigbie:
![screen shot 2017-01-07 at 5 31 34 pm](https://cloud.githubusercontent.com/assets/6854934/21745705/499b35ce-d4ff-11e6-8315-9cb28b16f191.png)
### Expected Behavior
When today's date is selected as starting date, then no data displays
### Actual Behavior
When today's date is selected as starting date, data displays
### Reproduction Steps

michaelmyers:
I can no longer reproduce, looks to have been fixed in #177.  Closing.

PR #177, opened 2017-01-10
Date query

### LINKED ISSUE #121: Better Querying By Date
Currently, we only allow you to look at the past seven days.  We want the ability to go back in time to as far back as they have logs.  

The primary route for searching back historically will be on the /logs page.  The user could either:

1. Set dates in the two date pickers and have it contact the logless server for the logs
2. Infinite scroll on the log list will automatically fetch the next set of logs (however this one might be a little more controversial as to if the user is expecting this behavior or not)

chrsdietz:
Fixes #121

- Add chai-datetime which allows for better date comparisons.
- User can not query logs all the way to the beginning of time.
- LogsPage now queries independently from the state of the app.  This means that:
- - On load, it will take the current default (last week).
- - The user can filter different dates. -> The logs page will query new logs.  These logs will *not* change the state of the website.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "browser-sync": "^2.17.5",
     "bs-fullscreen-message": "^1.1.0",
     "chai": "^3.5.0",
+    "chai-datetime": "^1.4.1",
     "chai-enzyme": "^0.6.1",
     "connect-history-api-fallback": "^1.3.0",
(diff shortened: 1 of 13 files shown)


chrsdietz:
I believe this PR will fix #170 as well.  I am not able to repeat it.

PR #194, opened 2017-01-30
Filter paging

chrsdietz:
- When filtering intent or log level, the system will automatically page new items until it fills up to 50 items or runs out.


src/pages/logspage/FilterableConversationList.tsx
```diff
@@ -44,14 +43,17 @@ export class FilterableConversationList extends React.Component<FilterableConver
...
-                me.state.shownConversations = items;
-                me.setState(me.state);
-            }).catch(function (err: Error) {
-                // only print out errors that are NOT the no items found error
-                if (err.message !== ERROR.FILTER.NO_ITEMS_FOUND) {
-                    console.error(err);
+            .then(function (result: FilterResult<Conversation>) {
+                if (result.changed) {
+                    let items = result.result;
+                    me.state.shownConversations = items;
```
michaelmyers: If we go to `then((result: FilterResult<Conversation>) => {` instead of needing to define `let me = this` outside the block.

PR #214, opened 2017-02-09
Tail

### LINKED ISSUE #184: Add tail function for logs
When looking at the logs for a particular skill, it should allow for auto-refresh.

This can be an option, such as adding a "LIVE" or "TAIL" toggle at the top.

With it, it should automatically add the latest logs at the top of the screen. Ideally, it should feel as close to live as possible.

chrsdietz:
Fixes #184 

- The Logs page will now auto-update every five seconds when on.
- The auto-update can be turned off via checkbox
- The auto-update will not be running when filtering from dates other than today.


michaelmyers:
Closing in favor of #216

PR #215, opened 2017-02-10
Integration

chrsdietz:
- Added an Integration page to the make panel to remind people how to use it.


chrsdietz:
I thought 32 files was a bit extreme, then I remembered that this includes #214 PR as well, so that one should be checked out first.

michaelmyers:
I apologize, I added this review to the wrong PR.

src/frames/Dashboard.tsx
```diff
@@ -139,6 +139,10 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
         {
           icon: "list",
           name: "logs"
+        },
+        {
+          icon: "description",
```
michaelmyers: Since it is code samples, how about `code` for the icon?

https://material.io/icons/#ic_code

PR #248, opened 2017-03-07
Google Home Summary Graphs

### LINKED ISSUE #232: Add Google Action Timeseries Event Data to Existing Alexa Data
screen shot 2017-02-22 at 6 06 31 pm
(Please excuse the extremely bad photoshop)

We want to display the Google Action data along side the Alexa data in the current time series events graph.

chrsdietz:
Fixes #232 

- Both the time series map and the intent series map will show data for both platforms now.


michaelmyers:
302 total events but only 117 from Alexa and 0 from Home.  People might think the 185 not accounted for are from another source but I assume this is because not all log events have information about the origin?

screen shot 2017-03-07 at 4 20 57 pm

chrsdietz:
This is a bug in the Logless server.  It's counting requests and responses, but there are more types of logs than that.

PR #259, opened 2017-03-16
Using action again

chrsdietz:
Now using `IntentName` as a fallback in Google conversations


michaelmyers:
Request a review from @jkelvie to make sure the issue is fixed.

PR #264, opened 2017-03-22
Name generating service

chrsdietz:
Uses new name generating service to generate unique names and secret keys.

Also, fixes issue where sources were not saving the date created.

### DIFF
--- a/src/services/source.ts
+++ b/src/services/source.ts
@@ -1,9 +1,9 @@
-import * as Firebase from "firebase";
+// import * as Firebase from "firebase";
 import "isomorphic-fetch";
 
 import { Query } from "../models/query";
 import { Source } from "../models/source";
-import StringUtil from "../utils/string";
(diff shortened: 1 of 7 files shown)


src/services/source.ts
```diff
@@ -1,78 +1,67 @@
-import * as Firebase from "firebase";
+// import * as Firebase from "firebase";
```
michaelmyers: Remove comments if no longer needed.

src/services/source.ts
```diff
@@ -1,78 +1,67 @@
...
 import "isomorphic-fetch";
 
+import { Query } from "../models/query";
 import { Source } from "../models/source";
-import StringUtil from "../utils/string";
+// import StringUtil from "../utils/string";
```
michaelmyers: Remove comments if no longer needed.

chrsdietz:
Comments removed

PR #272, opened 2017-04-06
Source url

### LINKED ISSUE #237: Create a new Source from URL
For fast & easy Source creation & onboarding, we want the ability to create a new source from the supplied query parameters in a URL.  

The URL would be of some format like:

```
https://bespoken.tools/dashboard?id=mighty-goose&code=<HASHED_KEY> 
```

To workflow needs to be as follows:

1.  If the user does not exist or is not logged in, the user will first be directed to the login/signup page, then directed to the dashboard
2. The dashboard receives the query parameters, determines if the skill exists for the user, if it doesn't it is created automatically.
3. The dashboard redirects you to the logs page of the skill.

chrsdietz:
Fixes #237

- There is now a redirect component that the page initially opens to.
 - - if queries "id" and "key" are not present, it instantly goes to "/skills"
 - - if queries "id" and "key" are present, it will try to link the source to the user.
 - - - If this succeeds, it will proceed directly to the source page.
 - - - If this fails, it will attempt to retrieve the source by ID and key.
 - - - - If ^that^ succeeds, it will go to the source page.
 - - - - if it fails, it will go to a 404 page.
 - Oh yeah, there's a new 404 page too.


src/pages/sourcepage/SourcePage.test.tsx
```diff
@@ -64,12 +64,14 @@ describe("Source Page", function () {
             expect(dataTiles).to.have.length(1);
         });
 
-        it("Tests that the source details header is gone when source is not defined.", function () {
+        it("Tests that nothign is displayed when source is not defined.", function () {
```
michaelmyers: `nothing` is misspelled

PR #283, opened 2017-04-26
Support for sharing conversations via URL

### LINKED ISSUE #163: Add ability to link to log
As of right now, you can link directly to a logs page like so:

`{host}/dashboard/skills/{sourceid}/logs`

It would be nice to look directly at a selected page via URL with parameters.  Like:

`{host}/dashboard/skills/{sourceid}/logs&id={convoId}`

This would allow collaborators to exchange the URL and go directly to it.

chris-ramon:
Do not merge yet!

Fixes #163

- [x] Extends logs actions & service to allow specify which endpoint to use.
- [x] Updates `ConvoPage` and `ConvoListPage` components to send downwards the current `Location` via context.
- [x] Exports `Location` interface from `createpage/Route.tsx`.
- [x] Updates `ConvoPage` and `ConvoListPage` unit tests, to handle dependencies on `context` and `Location`.
- [x] Appends queryparam `?id=convoId` when conversation is selected.

output

@chrsdietz


src/services/log.ts
```diff
@@ -50,10 +50,13 @@ namespace LogService {
     const BASE_URL = "https://logless.bespoken.tools/v1";
     // const BASE_URL = "https://logless-dev.bespoken.tools/v1";
 
-    export function getLogs(query: LogQuery): Promise<Log[]> {
-
+    export function getLogs(query: LogQuery, endpoint?: string): Promise<Log[]> {
         let url = BASE_URL + "/query?" + query.queryString;
 
+        if (endpoint) {
+          url = BASE_URL + endpoint;
```
jkelvie: Is this indented by a 4-space tab?

chris-ramon: good catch!, we might want to add fixed indentation to tslint configuration, still being discussed on tslint issue: https://github.com/palantir/tslint/issues/581

src/services/log.ts
```diff
@@ -50,9 +50,16 @@ namespace LogService {
...
+
+    export function getLogs(query: LogQuery, url?: string): Promise<Log[]> {
+        let defaultUrl = BASE_URL + "/query?" + query.queryString;
 
-        let url = BASE_URL + "/query?" + query.queryString;
+        if (url) {
+            defaultUrl = url;
+        }
 
         return fetchJson(url).then(function (json) {
```
jkelvie: This is using the URL parameter - so do we still need the defaultUrl?

jkelvie:
I only see the one log when I use this - I don't see the ones that surround it. Are we on the same page that we wanted to show the 25 before and after? Is that very hard to do?

chris-ramon:
> I only see the one log when I use this - I don't see the ones that surround it. Are we on the same page that we wanted to show the 25 before and after? Is that very hard to do?

Good catch! @jkelvie - we have the endpoint ready on `logless-server`, addressed on: https://github.com/bespoken/dashboard/pull/283/commits/3d225601ad376712b0b47b218d0882cf2357bc6d

Also I've just noticed 504 errors from `logless-server` when we query for transactions before/after, I'm taking a look to it.

PR #292, opened 2017-05-02
fixes for login page #282

### LINKED ISSUE #282: Differentiate UI between login/sign up pages
The login and registration pages are very similar - it is not completely clear which one the user is actually on.

Adding a title and potentially modifying the color and other UI elements would make it clearer to the user where they are.

(Based on feedback from @quentindela)

AnkRaiza:
Fixes #282 

- login ui
- login form test


chris-ramon:
LGTM :+1: 

Just would like to suggest naming `Log In` the header and button for consistency with URL and `Login with Github` button, as shown next:

image

jkelvie:
Good idea, though I would go with just "Login" instead of "Log In".

-John

src/components/AuthForm.test.tsx
```diff
@@ -165,10 +165,10 @@ describe("AuthForm", function () {
 
             let loginBtn = wrapper.find("Themed").first();
 
-            expect(loginBtn.prop("label")).to.equal("Login");
+            expect(loginBtn.prop("label")).to.equal("Log in");
```
jkelvie: Does this test pass?

PR #294, opened 2017-05-03
290 Extract text from response payload

### LINKED ISSUE #290: Extract text of response to user from response payload and show in conversational view
If the response has TTS (not SSML), it should extract it and put it in the the conversation bubble on the left-hand side.

The information will come from the response.outputSpeech.text element. If this element is not set, then nothing should be shown on the conversation bubble.

Alexa JSON reference is [here](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#response-body-syntax).

If the text of the output speech does not fit in the bubble on the left, it should be shortened to show as much as possible and then "..." at the end.

AnkRaiza:
Extract text of response to user from response payload and show in conversational view

Fixes #290 

- ConvoMainContent component
- ConvoMainContent unit test


src/models/conversation.ts
```diff
@@ -261,6 +266,12 @@ class AlexaConversation extends GenericConversation {
         return requestType;
     }
 
+    get outputSpeechText(): string | undefined {
```
jperata: this one seems to never return undefined

src/pages/logspage/list/ConvoMainContent.test.tsx
```diff
@@ -69,13 +83,29 @@ describe("ConvoMainContent", function () {
...
-                it("Tests that the conversation text is displayed.", function () {
+                it("Tests that the conversation subtitle text is displayed.", function () {
                     const intentWrapper = wrapper.find("span").at(1);
 
                     expect(intentWrapper).to.have.text(conversation.requestPayloadType);
                 });
             });
 
+            describe("outputSpeech text", function(){
+                it("Tests that the conversation response text doesn't get rendered when no text is avaible", function(){
```
jperata: typo available

chris-ramon:
lgtm :+1: — kudos on the coverage increase.

PR #295, opened 2017-05-03
fix 291 re arrenge conversation right hand side

### LINKED ISSUE #291: Re-arrange right-hand side of logging window
The response section should come before the logs.

Additionally, it would be nice if the request and response could have the titles fit inside their panels (say in the upper-right corner), so that the screen is not taken up to show "Request" and "Response" above the panels.

Please suggest some ideas on how we can make this show up better.

AnkRaiza:
Fixes #291 

- Interaction component
- outPutList component

BEFORE
screen shot 2017-05-03 at 2 54 17 pm

AFTER
screen shot 2017-05-03 at 3 40 43 pm


chris-ramon:
lgtm :+1:

PR #312, opened 2017-05-11
fix the undefined errors on get outputSpeech and create unit test

### LINKED ISSUE #313: Logs UI Freezes with Errors
### Expected Behavior

To be able to continuously scroll through the logs

### Actual Behavior
I start to scroll down and then hit a payload that causes errors in the console.

scroll-bug

This I just open the logs page and it errors.

scroll-bug-2

### Reproduction Steps

Most of these are occurring on Alexa Skills that use the AudioPlayer api.  Some I can get through a couple of logs before the console has errors, some don't load any.

AnkRaiza:
Fixes #313 

- replace getoutput speech function to validate each member of the object
- create unit test for undefined and regular log


chris-ramon:
lgtm :+1:

PR #320, opened 2017-05-16
add play button when there is audio file to play and show ssml text

### LINKED ISSUE #310: Add audio playback for interactions that include SSML or the AudioPlayer
This should show up as a play icon on the right-hand side of the interaction bubble.

[For SSML based responses](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/speech-synthesis-markup-language-ssml-reference), there may be multiple <audio> tags inside an SSML response. If there are multiple, take the first one.

Additionally, for SSML based responses, if there is a \<speak\> tag, take the content of that and show it as text output. Be careful about additional tags internally - they should be escaped (such as \<break\>, \<emphasis\>, etc.). \<audio\> tags within the SSML should be stripped of the URL for the sake of brevity. For example, an SSML like this:
`<speak>Hello there: <audio src="https://s3.amazonaws.com/xapp-alexa/BrandHaiku-HomePageStreaming-2892-PROMPT.mp3"></audio> </speak>`

Should be displayed as text like this:
`Hello there: <audio src="..." /> `

NOTE - we can strip out speak as well, since it is always there.

Additionally, for AudioPlayer playback (responses with AudioPlayer.Play directives in them), grab the URL from the audioItem.stream and use that for the audio. Here is an example payload:
```
{
  "version": "1.0",
  "response": {
    "shouldEndSession": true,
    "card": {
      "type": "Simple",
      "title": "Playing One",
      "content": "Playing One"
    },
    "directives": [
      {
        "type": "AudioPlayer.Play",
        "playBehavior": "REPLACE_ALL",
        "audioItem": {
          "stream": {
            "url": "https://d2mxb5cuq6ityb.cloudfront.net/Brand_Haiku-One-35ed7f07-f5ff-4d89-93ca-4621e5bcc4f6.m4a",
            "token": "0",
            "expectedPreviousToken": null,
            "offsetInMilliseconds": 0
          }
        }
      }
    ]
  }
```

AnkRaiza:
Fixes #310

-create audio component and basic render test
-change test to match new comboMainContent
screen shot 2017-05-16 at 4 23 29 pm
screen shot 2017-05-16 at 4 23 44 pm
screen shot 2017-05-16 at 4 23 57 pm


src/components/Audio/AudioComponent.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import * as React from "react";
+
+import AudioComponent from "./AudioComponent";
+
+const expect = chai.expect;
+
+describe("AudioComponent", function () {
+
+  it("renders properly", function () {
+    const wrapper: ShallowWrapper<any, any> = shallow(<AudioComponent src="https://s3.amazonaws.com/sounds226/boom.mp3" />);
```
chris-ramon: I think it would better to have `https://s3.amazonaws.com/sounds226/boom.mp3` as a constant, since its being used in several places.

src/models/conversation.ts
```diff
@@ -278,6 +282,36 @@ class AlexaConversation extends GenericConversation {
                 this.response.payload.response.outputSpeech.text;
     }
 
+    get ssmlText(): string | undefined {
+      const ssmlString = this.response &&
```
jperata: We are using this kind of validation a lot, if we extract an util function to validate it would make the code more readable.
This function
`function checkNested(obj /*, level1, level2, ... levelN*/) {
  var args = Array.prototype.slice.call(arguments, 1);

  for (var i = 0; i < args.length; i++) {
    if (!obj || !obj.hasOwnProperty(args[i])) {
      return false;
    }
    obj = obj[args[i]];
  }
  return true;
}
`
can check for it, based on [this stack overflow question](http://stackoverflow.com/questions/2631001/javascript-test-for-existence-of-nested-object-key).

src/utils/validation.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
jperata: Remove comment

src/utils/validation.test.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
jperata: Remove comment

PR #322, opened 2017-05-17
src: fixes on new skill creation

### LINKED ISSUE #321: Bug: failed to save new skills
### Expected Behavior
4. New skill is created

### Actual Behavior
4. Nothing happens, neither a error is shown.

### Reproduction Steps
1. Go to: `/dashboard/skills/new`
2. Write a skill name.
3. Click on create source.

chris-ramon:
Fixes #321

#### Overview
- [x] Adds missing `prop` for showing if new skill form fails.
- [x] Adds `services/source` unit test to make sure `SourceService.createSource` throws an error on failure.
- [x] Updates source service to handle `undefined` source's fields, this happens because Firebase API expect a valid JSON object when creating objects:
> The issue here is that 'undefined' is not a type in JSON, and it's not something that Firebase can store. 

> We *could* just treat an "undefined" as null and just delete / ignore it, but in our experience most of the time when there's an undefined in an object it's indicative of a bug in someone's code. In this case, they'd prefer to see an error rather than simply have their data disappear.

> Since we're a database, we try to err on the side of consistency / no data loss when these sorts of things come up.
https://groups.google.com/d/msg/firebase-talk/W2iw6L2Ghqo/fFmARTGifpQJ

#### Test plan
- [x] Make sure error message are shown when trying to create a new skill:

image


src/services/source.ts
```diff
@@ -86,14 +86,17 @@ export namespace source {
...
-                return sourcesPath.child(source.id).set(source)
+                return sourcesPath.child(source.id).set(JSON.stringify(source))
                     .then(function () {
                         // Save the source to the user's list of sources
                         return ref.child("users").child(user.uid).child("sources").child(source.id).set("owner");
                     });
             }).then(function () {
                 // And finally provide it back to the callback
                 return mutableSource;
+            }).catch((err: Error) => {
```
jperata: missing a test for this one, maybe you can generate a throw in the mocked db object?

chris-ramon: good call!, addressed on: 5cb653b3a25f7a6daf69200a4114abef200d5672

PR #337, opened 2017-05-25
v0.11.6

chris-ramon:
- Releases v0.11.6


dist/scripts/dashboard.js
```diff
@@ -78,7 +78,7 @@
 	var reducers_1 = __webpack_require__(1258);
 	var index_utils_1 = __webpack_require__(404);
 	var store_1 = __webpack_require__(1264);
-	console.log("v" + ("0.11.4") + "-" + (undefined));
+	console.log("v" + ("0.11.5") + "-" + (undefined));
```
jperata: shouldn't it vi 0.11.6 also?

chris-ramon: Yes, it should be `v0.11.6`; this file is auto-generated tho.

Let's updated it to 0.11.6 and see in next deployment whats going on.

Seems it's behind one version since later versions patches:
https://github.com/bespoken/dashboard/pull/334/files#diff-bfe5fb8713be78c76e14503b4f4c81d1R81

Issue #342, opened 2017-05-29
Source sorting is out of order


jkelvie:
### Expected Behavior
Sources in the source list should be sorted by name, case-insensitive

### Actual Behavior
It appears that the source list is being sorted by id rather than name, as newer sources where the name and id are not the same show up in seemingly random places in the list.

See the output below - "A Test Skill to Delete" is a new one that I added:
<img width="703" alt="screen shot 2017-05-29 at 3 07 51 pm" src="https://cloud.githubusercontent.com/assets/17267/26561028/b0ce1d42-4480-11e7-9e3b-cf39d73d5758.png">

### Reproduction Steps
Add sources, see that their names are out of order

chris-ramon:
Should it be sorted by created date?, because users might want to see latest skills created first in the list.

jkelvie:
I think anything other than alphabetical is likely to confuse.

-John

jkelvie:
#344

Issue #335, opened 2017-05-24
For SSML audio elements, display the last part of the filename


jkelvie:
Right now we show:
"<audio "..." />"

Instead, for an audio tag like:
`<audio src="https://audiohosting.com/my/path/MyFile.mp3" />`

We should show:
`<speak><audio src="//MyFile.mp3" /></speak>`

Additionally, please note for SSML, let's not show the quotes on either side, but instead the open and closing speak tag. I think that will look better, and will also clearly indicate that it is SSML.

jkelvie:
#344

PR #344, opened 2017-05-30
Fixes for source list order, ssml text format, login reset password button new alignment

### LINKED ISSUE #343: Place "Reset Password" link underneath the sign-in box
It should go below the white area, to make the buttons seem less crowded.

Additionally, I would remove two of the lines:
1) The one above Register and Login
2) The one above Login with Github

Lastly, I would give all the buttons a background color, even when not selected. A blue similar to the color of the reset password link could work.

screen shot 2017-05-29 at 3 11 27 pm

AnkRaiza:
Fixes #343 , #342 , #335 

-conversation.ts
-conversational buble
-source list order
-login page


src/pages/SourceListPage.tsx
```diff
@@ -34,10 +34,15 @@ function mapStateToProps(state: State.All) {
 export class SourceListPage extends React.Component<SourceListPageProps, SourceListPageState> {
 
     render() {
+        const sortedList = this.props.sources.sort((a, b) => {
```
jperata: since string already manage comparison operators , you can just do, sources.sort() , it will sort in asc order, if you need desc you can use revert

src/models/conversation.ts
```diff
@@ -282,8 +282,10 @@ class AlexaConversation extends GenericConversation {
 
     get ssmlText(): string | undefined {
       const ssmlString = propertyExist(this.response, "payload", "response", "outputSpeech", "ssml") ? this.response.payload.response.outputSpeech.ssml : undefined;
-      const removedSpeak = ssmlString ? ssmlString.replace("<speak>", "").replace("</speak>", "").trim() : undefined;
-      return removedSpeak ? removedSpeak.replace(/src=".*?"|src='.*?'/g, '"..."') : undefined;
+      const audioUrls = ssmlString ? ssmlString.match(/src=".*?"|src='.*?'/) : undefined;
```
jperata: why not
const audioUrls = ssmlString && ssmlString.match(/src=".*?"|src='.*?'/) ; 
same for the rest

jperata: If you need the explicit undefined for some reason you can just return it at the end

src/pages/SourceListPage.tsx
AnkRaiza: this is an array of objects so i have to map to the correct property to sort also, using just sort() wont fix the case insensitive issue

PR #346, opened 2017-06-02
330 add response time graph

### LINKED ISSUE #330: Add response rate graph
We want to add a response rate graph that shows the average time in milliseconds that it takes for a request/response to be processed.

This data will be made available on the logless-server via an endpoint to be created here:
https://github.com/bespoken/logless-server/issues/45

The data should initially be presented for the entire trailing week, similar to other graphs on the summary screen.

The graph should be the first one the user sees when they open the summary screen.

AnkRaiza:
Fixes #330

- add response time graph


src/pages/sourcepage/SourceResponseTimeAverage.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const currentDate: moment.Moment = start.startOf("day");
+  while (currentDate.isBefore(end.endOf("day"))) {
+    intervalArray.push({interval: moment(currentDate).startOf("day").valueOf(), avgResponseTime: 0, intervalDate: moment(currentDate)});
+    currentDate.add(interval, "minutes");
+  }
+  return intervalArray;
+}
+
+function formatIntervalDate(stringToReplace: string, intervalValue: number) {
+  const replacedFormat = stringToReplace.replace(/-/g, function(match, i) {
```
jperata: i think using moment functions read a little better 
moment(stringToReplace, "YYYY-MM-DD-hh-mm").format("YYYY-MM-DD hh:mm")

src/pages/sourcepage/SourceResponseTimeAverage.tsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+interface SourceResponseTimeAverageState extends LoadingComponent.LoadingComponentState<IntervalData[]> {
+  intervalArray: IntervalData[];
+}
+
+export class SourceResponseTimeAverage extends LoadingComponent.Component<IntervalData[], SourceResponseTimeAverageProps, SourceResponseTimeAverageState> {
+
+    static defaultProps: SourceResponseTimeAverageProps = {
+        source: undefined,
```
jperata: isn't this the same of not adding source at all?

src/components/Graphs/Line/IntervalChart.tsx
```diff
@@ -0,0 +1,101 @@
...
+        dot: false,
+        type: "monotone"
+    };
+
+    static createTicks(props: IntervalChartProps): number[] {
+      const data: IntervalData[] = props.data;
+      if (data.length === 0) {
+        return [];
+      }
+      let highest: moment.Moment = moment(data[0].intervalDate).startOf("day");
```
AnkRaiza: this si just for that ticks that will shot month day combination so no data will be actually omited.

src/pages/sourcepage/SourceResponseTimeAverage.tsx
AnkRaiza: This is on order to match the SourceResponseTimeAverageProps

PR #354, opened 2017-06-14
src: link source fixes

chris-ramon:
#### Overview
- [x] Fixes linking source.
- [x] Refactors linking source code.
- [x] Removes `createpage/Route.tsx` component:
  - By extracting linking source code to the `Dashboard` component, to keep it more simple and consistent.
  - All the code, can be summarized within `Dashboard.ComponentDidMount`, the interesting part: https://github.com/bespoken/dashboard/compare/linking-source-refactor?expand=1#diff-e0d107de60f5989c4108f2fc3899c3b0R100

#### Test plan
- [x] Unit test
- [x] Manual test:
  - Make sure linking source works as expected.
  - Make sure there is not more dependency on `createpage/Route.tsx`, as redirection extra step.


src/frames/Dashboard.tsx
```diff
@@ -94,7 +97,25 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
     return classNames(CLASSES.COLOR.GREY_100, CLASSES.TEXT.GREY_600);
   }
 
-  componentWillMount() {
+  async componentDidMount() {
```
jperata: Be aware that when you do an async componentDidMount (or a componentWillMount) the render event won't await until the promise is solved.
It will render, let the function complete and if there's any change of state in the delayed work then it will render again.
This just seems to redirect as soon as it finish but you may need to have a loading state that you can update at the end of the function and hide the page until it's ready.

chris-ramon: good call @jperata; we are fine, having the `Dashboard` component being rendered at first; then re-rendered when new state is ready.

PR #355, opened 2017-06-19
src: fixes linking source redirection

chris-ramon:
- Improvements on `Dashboard` component, to redirect after sources are resolved.
- Make sure getSources action is promise based.


src/actions/source.ts
```diff
@@ -65,30 +66,33 @@ export function deleteSource(source: Source): (dispatch: Redux.Dispatch<any>) =>
...
+                    for (const k of keys) {
+                        promises.push(Bluebird.resolve(service.getSource(k)));
+                    }
+                }
+                Bluebird.all(promises.map((p) => p.reflect()))
+                    .each((inspection: Bluebird.Inspection<any>) => {
+                        if (inspection.isFulfilled()) {
+                            sources.push(new Source(inspection.value().val()));
+                        } else {
+                            console.log(inspection.reason());
```
AnkRaiza: are we gonna leave this log here?

src/frames/Dashboard.tsx
```diff
@@ -99,24 +99,22 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
         if (!pipe.diagnosticsKey) {
           await SpokeService.savePipe(self.props.user, source, pipe.http, true);
         }
-        redirect();
-      } catch (err) {
-        redirect();
-      }
-    } else {
-      this.props.goTo("/skills");
+      } catch (err) {}
```
AnkRaiza: isn't this generating an unused variable error?

src/index.tsx
```diff
@@ -95,14 +96,31 @@ Firebase.auth().onAuthStateChanged(function (user: Firebase.User) {
  *
  * See below on the onEnter method.
  */
-let checkAuth: EnterHook = function (nextState: RouterState, replace: RedirectFunction) {
+let checkAuth = function (nextState: RouterState, replace: RedirectFunction): boolean {
     const session: any = store.getState().session;
     if (!session.user) {
```
AnkRaiza: maybe just
!session.user && replace({...});
return !!session.user;

src/index.tsx
```diff
@@ -95,14 +96,31 @@ Firebase.auth().onAuthStateChanged(function (user: Firebase.User) {
...
             query: nextState.location.query,
             state: { nextPathName: nextState.location.pathname, query: nextState.location.query }
         });
+        return false;
+    }
+    return true;
+};
+
+let onEnterDashboard: EnterHook = function (nextState: RouterState, replace: RedirectFunction) {
+    if (!checkAuth(nextState, replace)) {
```
AnkRaiza: no need for brackets here if(!chekauth()) return;

AnkRaiza:
minor comments

src/actions/source.ts
chris-ramon: yes, we need to log, when something went wrong while monitoring a source.

src/frames/Dashboard.tsx
chris-ramon: it is a unused variable, but we should keep it to complain with tslint

src/index.tsx
chris-ramon: Interesting suggestion, but I think it reads better as it is

src/index.tsx
chris-ramon: you mean, curly braces; good call, addressed on: https://github.com/bespoken/dashboard/pull/355/commits/e4099d42a373f58654daee8ee05e0b682cc1658e

PR #359, opened 2017-06-20
fix save on lambda and hide aws credentials

### LINKED ISSUE #357: Hide AWS Credentials on Settings Page
These should be obscured, except for the last four-characters. The user can only update them, they cannot view them.

So after initially saving, all they will see is:
Access Key: ****************WDUQ
Secret Key: ***************************************49wu

The reason they are view-only is these are very sensitive and we don't want it to be easy for people to see them.

AnkRaiza:
Fixes #357 

- save lambda on settings page
- test
- integration spoke swapper


src/pages/integration/IntegrationSpokes.tsx
```diff
@@ -336,3 +368,8 @@ function validateUrl(check?: string): boolean {
     const regex = /^(https?:\/\/).+/;
     return regex.test(check);
 }
+
+function obscureInput(value: string): string {
+    const baseMask = "******************************************************************************";
+    return baseMask.substr(0, value.length - 4) + value.substr(-4, 4);
```
jperata: @chris-ramon recommended a cleaner way, avoiding baseMask:
`         return "*".repeat(stringToMask.length - 4) + stringToMask.slice(-4); `
performance it's not affected.

PR #361, opened 2017-06-21
add copy feature using titles

### LINKED ISSUE #352: Ability to Copy Request or Response Payloads
I would love to quickly copy the payloads so I can then use them for testing.  Something similar to a copy to clipboard button next to the payload would be awesome.

AnkRaiza:
Fixes #352 

-add copy feature to interaction page

screen shot 2017-06-21 at 4 09 03 pm


src/components/Interaction.tsx
```diff
@@ -98,6 +118,40 @@ export class Interaction extends React.Component<InteractionProps, InteractionSt
         return key;
     }
 
+    handleCopyRequestPayload: (e: any) => void;
+    handleCopyResponsePayload: (e: any) => void;
+    handleCopyPayload(e: any) {
+        const payload = e.target.getAttribute("data-type") === "request" ? this.props.request : this.props.response;
+        const textArea = document.createElement("textarea");
```
chris-ramon: Perhaps we should prefer object literal here.
```typescript
textArea.style = {
  position: "fixed",
  top: 0,
  left: 0,
  // ...
};
```

chris-ramon:
also perhaps, we want to use another icon?, such as:
Github:
image

Font Awesome:
image

/cc: @jkelvie

src/components/Interaction.tsx
AnkRaiza: style is a readonly property in this context, so i can't actually replace it with object literal, i can only set their properties values

chris-ramon: ok cool, didn't know that :+1:

PR #377, opened 2017-06-26
remove redirect on componentDidMount

### LINKED ISSUE #374: When I refresh, I lose the the section I'm on
Clicking refresh while looking at the logs, I end up being taken back to the skill list page.

This is a serious issue, as refreshing is a common thing to do on the site.

This appears to be a regression issue, as this works in the current production version.

AnkRaiza:
Fixes #374 

- dashboard
- index

#### Test plan
- Manual test:
  - [ ] Refreshing on `skills/:id` preserves the page.
  - [ ] Source linking works, and redirects to `skills/:id`.
  - [ ] Refreshing on `skills` preserves the page.


src/frames/Dashboard.tsx
```diff
@@ -100,21 +100,24 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+          if (!pipe.diagnosticsKey) {
+              await SpokeService.savePipe(self.props.user, source, pipe.http, true);
+          }
+          redirectTo();
+      } catch (err) {
+          redirectTo();
+      }
     }
-    this.props.getSources().then(() => redirectTo()).catch((err) => console.log(err));
+    this.props.getSources();
```
chris-ramon: @AnkRaiza we want to do the redirection, after the `getSources` is resolved, otherwise, 404 pages will be shown, since the source being shown for example on `skills/:id` won't be found in the global state.

AnkRaiza: i'm investigating this further, the 404 shows for me even with this redirect, right now the id and key gets lost always and we end up going to skills page on every page refresh (not espected behabior)

chris-ramon: I think the solution is to redirect to whatever the `location.path` is, doing this way will let us to the regression where we constantly see 404 pages.

jkelvie:
Where are we on this? I would like to get this one resolved

chris-ramon:
Manually tested the following:

- Success:
  - Source linking works, and redirects to skills/:id.
  - Refreshing on skills preserves the page.

- Fails:
  - Refreshing on skills/:id preserves the page.

jkelvie:
These changes have been overwritten by subsequent builds. I tested these scenarios previously and they all worked for me. I am merging, and we can test further along with other changes.

PR #380, opened 2017-06-28
add a way to differentiate github login from sign up

### LINKED ISSUE #362: Need way to detect when a user signs in for the first time
Either via firing off a Google Analytics event, or directing them to a special URL (such as adding a "#registered" to the their source). Without this, we cannot see the complete conversion funnel.

Firing a google analytics event is as easy as adding a line like this:
`ga("send", "event", "Registered", "Registered")`

We nee to make sure Google Analytics Javascript library is already installed, but I believe it is.

AnkRaiza:
Fixes #362 

- auth
-  auth.test


src/services/auth.ts
```diff
@@ -11,32 +11,34 @@ import { remoteservice } from "./remote-service";
...
-        let firstPromise: Promise<User>;
+    async function loginWithProvider(provider: remoteservice.auth.AuthProvider, auth: remoteservice.auth.Auth = remoteservice.defaultService().auth(), storage?: LocalStorage, db: remoteservice.database.Database = remoteservice.defaultService().database()): Promise<User> {
+        let result: any;
+        let ref = db.ref();
         if (browser.isMobileOrTablet()) {
-            firstPromise = auth.signInWithRedirect(provider);
+            result = await auth.signInWithRedirect(provider);
         } else {
-            firstPromise = auth.signInWithPopup(provider);
+            result = await auth.signInWithPopup(provider);
```
chris-ramon: should we add try/catch to handle errors?

PR #399, opened 2017-07-11
add bback transactions before and after

### LINKED ISSUE #305: Debug performance issues with loading transactions before and after when opening by transaction ID
The calls to Logless seem to timeout when getting the 25 transactions before and after the designated one.

This is likely an issue with Mongo and logless server, but recording here as this is where the behavior manifests.

For now, we just load the single transaction itself.

AnkRaiza:
Fixes #305

- partially fixes 305 we need the logless server pr as well


src/services/log.ts
```diff
@@ -52,17 +52,27 @@ namespace LogService {
...
+            transactionUrl = transactionUrl + "&transactions_after=" + transactionsAfter;
+        }
+        return transactionUrl;
     }
 
     export function getLogs(query: LogQuery, url?: string): Promise<Log[]> {
         if (!url) {
             url = BASE_URL + "/query?" + query.queryString;
         }
+        console.log(url);
```
jperata: remove

AnkRaiza:
i was getting timeouts when not validating the logs.length and not having either after or before transactions, after the validation change i didn't receive any timeout

PR #407, opened 2017-08-02
add ssml and audio url for google plus some actionsSdk support

AnkRaiza:
Fixes #google support


src/models/conversation.ts
```diff
@@ -380,6 +388,42 @@ class GoogleHomeConversation extends GenericConversation {
         return this.intent;
     }
 
+    get outputSpeechText(): string | undefined {
+        let responseText: string | undefined;
+        if (propertyExist(this.response, "payload", "expect_user_response")) {
+            if (this.response.payload.expect_user_response) {
+                responseText = ""; // work in progress
```
jkelvie: ?

src/models/conversation.ts
```diff
@@ -367,6 +367,14 @@ class GoogleHomeConversation extends GenericConversation {
                         }
                     }
                 }
+
+                if (payload.inputs) {
+                    const data = payload;
+                    if (data.inputs && data.inputs.length > 0) {
```
jkelvie: Can this logic and the logic above be consolidated? Once we have the data object, seems like it is the same between the two.

PR #414, opened 2017-08-16
click on item and click on icon to work properly

### LINKED ISSUE #412: After filtering logs by user (by clicking on the device icon), cannot return to default view
### Expected Behavior
When I click to filter by user, by clicking on the little device icon, I get the only the logs for that user.
When I unclick it, I should revert to seeing all logs in the default view.

### Actual Behavior
When I click to filter by user, by clicking on the little device icon, I get the only the logs for that user.
When I unclick it, I get all the recent logs around the one that I clicked on.

The root cause is that the URL gets updated to the ID for the particular log the user clicked on the device of. When they unclick, the URL should revert to the plain URL.

Additionally, I think if the user clicks again on a log, it should de-select it and also revert the URL (which is another annoyance).

AnkRaiza:
Fixes #412 

- ConvoList
- ComboExplorer


src/pages/logspage/list/ConvoList.test.tsx
```diff
@@ -160,7 +160,8 @@ describe("ConvoList", function () {
                 expect(listItem.props["active"]).to.be.false;
             });
 
-            it("renders two interactions after clicking on two list items", function () {
+            // Do we actually support more than one selected item? i haven't seen that behabior on the dashboard
```
jkelvie: We do not

src/pages/logspage/ConvoExplorerPage.tsx
```diff
@@ -45,6 +45,12 @@ export class ConvoExplorerPage extends React.Component<ConvoExplorerPageProps, C
             };
             location.query = {...location.query, ...query};
             browserHistory.push(location);
+        } else if (browserHistory) {
+            this.state.selectedConvo = undefined;
+            this.setState(this.state);
```
jperata: ` this.setState({selectedConvo: undefined});`
 is cleaner and the recommended way (you shouldn't touch this.state by setting values)

PR #435, opened 2017-09-06
average response time graph improvements

chris-ramon:
#### Overview
- This PR adds performance improvements to the average response time graph.
- [x] `SourceResponseTimeAverage`: 
  - Removes data parsing, now it lives in the endpoint.
  - Updates to simpler props.
- [x] `IntervalChart`:
  - Removes ticks creation, delegated to the endpoint.
  - Now receives ticks.
- [x] LOGLESS_BASE env variable:
  - As default points to prod.

#### Test plan
- [x] Manual test


jkelvie:
Does this include the parameterization for which logless instance to point at?

PR #461, opened 2017-10-09
multi user support

### LINKED ISSUE #424: Enable multi-user support
Per this spec:
https://docs.google.com/a/bespoken.tools/presentation/d/169Xzz4ZYHEXxZ3phyZXCwqHQ3lYlJpa-Hxe5E5mqBmI/edit?usp=sharing

chris-ramon:
Fixes #424

- User
- Source
- Team page


jkelvie:
I pushed the build back out - here are my initial testing notes:

- Menu does not close after I click on Team (the team page comes up, but the menu is still open - see screenshot)
- Font at the top of the page is too large, and should have more margin from the top (maybe 10px more?)
- Error on adding a member (see screenshot)

Can we used Khand as the standard for page titles like this?

Screenshots below.
screen shot 2017-10-11 at 11 54 11 am
screen shot 2017-10-11 at 11 53 54 am

jkelvie:
On the add user page:
1) The email field is labeled name
2) If I put in a non-email address, there is no validation error
Instead, I just cannot click add
3) If I put in an email that does not existing in the system, there is no error
But I cannot add them either

jkelvie:
When I logged in with the user I added, I only had one skill visible - it should be all the skills for the user that granted me access (I logged in as jpk@bespoken.io)

jkelvie:
I get a 404 page after adding a user - when I go and check, the user is added, but I land on a 404 page immediately afterwards

jkelvie:
Users that are viewers should not be able to view the secret key (there should not a be a "show" button next to it)

jperata:
Closing this PR, the branch is still there so we can grab the changes from there.

Issue #464, opened 2017-10-09
Use SMAPI integration to validate user is owner of skill


jkelvie:
We can use the Skill Management API to verify which skills a user is associated with, and limit them to invoking skills which they are the owner of:
https://developer.amazon.com/docs/smapi/ask-cli-intro.html#smapi-intro

To implement this, we need to add:
1) Add additional scopes to token authorization process
For getting the token on the validation page, we want to add the scope:
alexa::ask:skills:read
alexa::ask:models:read

The user will be granted a token that has this additional privilege.

For users that have older tokens, they will get an error message when they run the new validation page that they need to refresh their token.

2) Scrape the SMAPI to get the skills the user manages
This should be done inside the source-api, when it gets the new token from the silent echo service.

The data from the calls should be captured in the Firebase DB and associated with the user.

Each skill associated with the user should be stored as a "skills" record, with the following fields:
id - Skill ID
interactionModel - The full interaction model
skillManifest - The skill manifest
userID - The user ID that imported this skill (a skill may still be associated with multiple users)

Additionally, the user record on users should contain a "skills" array that contains an array of skill IDs associated with the user.

3) Add an endpoint to source-api to check if a token has access to a particular skill
/skill_authorized
The endpoint will take an invocation name, and will return plain text with either:
AUTHORIZED
NOT_AUTHORIZED
UNKNOWN

In the case of unknown, it means the user has not granted the correct privileges to access the skill. They should be shown a message saying: "Security token lacks sufficient information. Please re-authenticate with Amazon here to update your security token."

4) The silent echo sdk should call the new /skill_authorized endpoint with the invocation name of the skill before running each sequence. If it is not authorized, a message should be shown to the user.

jkelvie:
We will also need to capture the vendor ID for the user on the validation page - this is a required parameter to make the list skills call. The vendor ID for a user can be found here:
https://developer.amazon.com/mycid.html

We should add a field for this, as well as instructions for how to retrieve it.

The field should go above the token, and say below it:
To retrieve your vendor ID, [click here](https://developer.amazon.com/mycid.html). Please make sure it is for the correct organization if you belong to multiple.

jkelvie:
From initial review:

1) The vendor ID and token take up too much space on the page - can we make these sections smaller by:
A) Reducing font size? 
B) Reducing spacing between the fields and the text below them
Overall, these fields should consume much less screen real estate

2) The link to get the vendor ID should open a new tab/page.

[...]

Issue #465, opened 2017-10-09
prod - dev matching assets path


chris-ramon:
- within harness directory, let's create new assets directory which should contain https://github.com/bespoken/dashboard/blob/master/harness/images/Schedule.svg and https://github.com/bespoken/dashboard/blob/master/harness/images/Spinner.svg — so we match the path being use for assets in production

PR #466, opened 2017-10-10
ValidationPage: SMAPI integration

chris-ramon:
#### Overview
- ValidationPage:
  - [x] Adds & wires `Vendor ID` ui.
  - [x] Finds and sends `smAPIAccessToken` when validating scripts.
  - [x] Saves `vendorID` to firebase.
  - [x] Loads and sets existing `vendorID`.
- Closes https://github.com/bespoken/dashboard/issues/464 & https://github.com/bespoken/dashboard/issues/465

output

#### Test plan
- [x] Manual test
  - [x] Vendor ID is save and set on validation page load: <details><summary>GIF</summary></details>
  - [x] Vendor ID if invalid for SMAPI, shows proper error message:<details><summary>GIF</summary></details>
  - [x] Form, for unknown errors, shows proper message error:<details><summary>GIF</summary></details>

#### Action required
- Merge along `source-api` PR: https://github.com/bespoken/source-api/pull/33


src/models/user.ts
```diff
@@ -1,7 +1,9 @@
 export class UserDetails {
     readonly silentEchoToken: string;
-    constructor(silentEchoToken: string) {
+    readonly smAPIAccessToken: string;
+    constructor(silentEchoToken: string, smAPIAccessToken: string) {
         this.silentEchoToken = silentEchoToken;
+        this.smAPIAccessToken = smAPIAccessToken;
```
jkelvie: We are essentially going to deprecate the silent echo token, correct? New users will not need both.

jkelvie: I would also call it SMAPIAccessToken - consistent with how Amazon does it.

chris-ramon: > We are essentially going to deprecate the silent echo token, correct? New users will not need both.

Well, we can preserve both, they clearly different tokens, one for virtual device, one for SMAPI.

Also preserving both will allow us to do no more changes on silent-echo side for SMAPI integration, more than the `/refresh_token` endpoint which is working great.

> I would also call it SMAPIAccessToken - consistent with how Amazon does it.

Good call!, I'll rename it.

jkelvie: They are actually the same token - the one token covers both sets of privileges. That's why I was saying we don't need both.

chris-ramon:
> Can we put the vendor ID and token fields on the same line next to each other? They are taking up too much space now

Good call!, addressed on: https://github.com/bespoken/dashboard/pull/466/commits/0bca9da5ee59c66937f68f0a251b25ce9b8def68


image

Issue #450, opened 2017-10-03
AudioPlayer Metrics


jkelvie:
Aggregate and display audio player metrics. These should include:

- Number of sessions
- Average session duration

To calculate a an audioplayer session, we need to create a derived collection in Mongo. I propose we use the following algorithm:
1. Look at each user in the logs
1. Look for an AudioPlayer.Play directive
1. Look for a corresponding AudioPlayer.PlaybackStopped directive or AudioPlayer.PlaybackFinished directive
1. Subtract the time on offsetInMilliseconds for Stopped or Finished from the offsetInMilliseconds on Started
1. Add this to running counter of session duration
1. Look for either a AudioPlayer.PlaybackStarted request, AudioPlayer.Stop, or AudioPlayer.Play directive response
   * If stop, the the session is completed - add a record to the AudioPlayerSession collection for the user
   * If playback started, repeat steps 3 to 7, adding the time for listening to the session counter
   * AudioPlayer.Play, add a record for the session then repeat steps 3 to 7 - for a new session

PR #471, opened 2017-10-31
audio session chart

AnkRaiza:
Relates #450 

- New audio player section and chart

screen shot 2017-11-03 at 4 40 32 pm


jkelvie:
For the chart,the time should be translated into minutes and seconds on the y-axis. The chart should be titled - "Average Session Duration".

I would put average duration and number of sessions as fields above the chart (with the field name and the value to the right of it). Average duration should also be translated into minutes and seconds here.

There should be an additional chart for number of sessions per day. This should be title "Sessions Per Day".

jkelvie:
This is looking close - I mean for the overall numbers for the period to be pulled out, similar to the summary stats on the main page:
screen shot 2017-11-03 at 11 04 08 am

So it will be Average Duration: 4:00 Number of Sessions: 500.

These numbers will be for the entire period.

jkelvie:
The chart looks good to me. I think this is ready to go functionally - @chris-ramon any comments on the code?

jkelvie:
Initial notes:
1) Time period resets to zero when it goes over an hour - instead it should show "1h:10m:15s" for a period of "70:15" - you can see this below on the y-axis for average session duration

2) The x-axis goes from most recent to oldest - it should be the other way around.

screen shot 2017-11-21 at 3 06 04 pm

src/pages/audioplayerpage/AudioPlayerStats.tsx
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    async startLoading(props: AudioPlayerStatsProps): Promise<LogService.AudioPlayerStats> {
+        const query: Query = new Query();
+        query.add(new SourceParameter(props.source));
+        query.add(new StartTimeParameter(props.startDate));
+        query.add(new EndTimeParameter(props.endDate));
+        const [sessions, duration] = await Promise.all([LogService.getAudioSessions(query), LogService.getAudioDuration(query)]);
+        const audioSessions = sessions.map((audioSession: any) => {
```
chris-ramon: @AnkRaiza could we move this operations: `sessions.map` and `audioSessions.reduce` to the API side.

src/pages/audioplayerpage/NumberSessions.tsx
```diff
@@ -0,0 +1,150 @@
...
+        if (!source) return;
+        this.mapState({ data: [], sourceState: 1});
+        const query: Query = new Query();
+        query.add(new SourceParameter(source));
+        query.add(new StartTimeParameter(this.props.startDate));
+        query.add(new EndTimeParameter(this.props.endDate));
+
+        let formatedData;
+        try {
+            const data = await AudioPlayerService.getAudioSessions(query);
```
chris-ramon: Also this part could we move it to the API side ?

src/pages/audioplayerpage/NumberSessions.tsx
```diff
@@ -0,0 +1,150 @@
...
+            return [];
+        }
+
+        const query: Query = new Query();
+        query.add(new SourceParameter(source));
+        query.add(new StartTimeParameter(startDate));
+        query.add(new EndTimeParameter(endDate));
+        try {
+            const data = await AudioPlayerService.getAudioSessions(query);
+            const audioSessions = data.map((audioSession: any) => {
```
chris-ramon: same here, could we move it to the API side ?

src/pages/audioplayerpage/SessionDuration.tsx
```diff
@@ -0,0 +1,145 @@
...
+        if (!source) return;
+        this.mapState({ data: [], sourceState: 1});
+        const query: Query = new Query();
+        query.add(new SourceParameter(source));
+        query.add(new StartTimeParameter(this.props.startDate));
+        query.add(new EndTimeParameter(this.props.endDate));
+
+        let formatedData;
+        try {
+            const duration = await AudioPlayerService.getAudioDuration(query);
```
chris-ramon: Also here, could we move it to the API side, as well as the operations within `startLoading`.

Issue #469, opened 2017-10-30
Integration page should include secret key in the code snippets


jkelvie:
### Expected Behavior
The users actual secret key should be integrated into code snippets

### Actual Behavior
The integration page, across all tabs, says <SECRET_KEY> instead of using the real secret key.

This prevents users from just cutting and pasting the sample code into their app.

The tabs for NodeJS, Express, Java and Cloud Functions should all be updated.

PR #473, opened 2017-11-03
The property showSecret was set to true on Integration Page

dmarvp:
Relates #469 

- the property showSecret was set to true in StateIntegrationPage.tsx, this shows the secret key in all the code snippets of the Integration page.


dmarvp:
@jkelvie all those pages already had tests, but they were only testing that "something" was being rendered. I changed it so that it does check that a secret key is rendered or not. And also made sure to cover all the cases:
- To have a secret key and tell the page to show it.
- To have a secret key and tell the page to not show it.
- To not have a secret key but still tell the page to show a secret key.
- To not have a secret key and tell the page to not show it.

chris-ramon:
lgtm :+1:

Issue #474, opened 2017-11-06
Enable email confirmation for email registrations


jkelvie:
When the user signs up via email (as opposed to via Github), they should be asked to confirm their email.

It seems this is a builtin feature of firebase - more information here:
https://firebase.googleblog.com/2017/02/email-verification-in-firebase-auth.html

When a user signs up, we should:
1) Automatically send them a verification email
The email should say:
```
Hello %DISPLAY_NAME%,

Follow this link to verify your email address:
https://apps.bespoken.io/dashboard/_/auth/action?mode=<action>&oobCode=<code>

If you didn't ask to verify this email address, you can ignore this email.

Thank you,
The Bespoken Team
```

The email should come from support@bespoken.io. The link used above is a variation on the one that the firebase blog uses, though with our domain name.

2) Once the email is sent, a toast message should be shown that says "Email verification sent"

3) When they re-enter the app, if they are not verified, as message should be shown as toast that says:
```
Your email is not yet verified - please click on the link in the email we sent to you at signup. Or click here to receive another verification email.
```

PR #476, opened 2017-11-11
Email verifications at signup and login

dmarvp:
Relates #474 

- Sending a verification email when signing up with email and password
- Warning users that haven't been verified and letting them receive another verification email
- Toast component


AnkRaiza:
check Juan's comment

PR #504, opened 2017-11-29
verification emails

dmarvp:
relates #474 

- A verification email is sent when a new user signs up using email and password
- When entering the dashboard, the user will get a notification if the email hasn't been verified yet
- The notification won't show up again until the user logs in again
- Github users won't see any notification


src/frames/Dashboard.tsx
```diff
@@ -85,15 +95,28 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
     this.handleSelectedSource = this.handleSelectedSource.bind(this);
     this.handlePageSwap = this.handlePageSwap.bind(this);
     this.handleHomeClick = this.handleHomeClick.bind(this);
     this.handleOpenModal = this.handleOpenModal.bind(this);
     this.handleCloseModal = this.handleCloseModal.bind(this);
     this.handleEnterContest = this.handleEnterContest.bind(this);
+    this.handleVerifyEmailClick = this.handleVerifyEmailClick.bind(this);
+    this.handleSnackbarHiding = this.handleSnackbarHiding.bind(this);
+
+    const snackBarLabel = (
```
chris-ramon: Since it is a constant, let's have it as static variable before the `constructor` method, like:
```typescript
static readonly snackBarLabel = (
  // ...
);
```
Also better to rename it to something more specific to its purpose, like: `EmailVerificationSnackbarLabel`.

src/frames/Dashboard.tsx
```diff
@@ -216,19 +247,29 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
 
   handleEnterContest() {
     window.open("https://www.surveymonkey.com/r/X5R3W8G", "_blank");
     window.localStorage.setItem("contest", "true");
-    this.setState({ showModal: false });
+    this.setState({ ...this.state, showModal: false });
+  }
+
+  handleVerifyEmailClick() {
+    remoteservice.defaultService().auth().currentUser.sendEmailVerification();
```
chris-ramon: Here, since [`sendEmailVerification`](https://firebase.google.com/docs/reference/js/firebase.User#sendEmailVerification) is a promise, we want to call line 267 within then callback, like:
```typescript
remoteservice.defaultService().auth().currentUser.sendEmailVerification()
  .then(() => {
    this.setState({ ...this.state, showSnackbar: "sent" });
  });
```

src/frames/Dashboard.tsx
```diff
@@ -18,6 +18,10 @@ import SpokeService from "../services/spokes";
 import ArrayUtils from "../utils/array";
 import { Location } from "../utils/Location";
 
+import Snackbar from "react-toolbox/lib/snackbar";
+import { remoteservice } from "../services/remote-service";
+
```
jperata: minor, avoid spaces between imports

src/frames/Dashboard.tsx
```diff
@@ -260,6 +301,19 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
         </Header>
         <Content>
           {this.props.children}
+          <Snackbar className="sm-snackbar" action="Dismiss" type="cancel"
```
chris-ramon: Here, I think we should simplify by not having `this.state.showSnackbar`, with different states: `verify`. `sent` and false,
instead lets use the raw data to control the state of which snackbar to show, like:

```typescript
{!this.props.user.emailVerified && !this.state.showEmailVerificationSentMessage
  ? (<Snackbar className="sm-snackbar" action="Dismiss" type="cancel"
    active={true}
    onClick={this.handleSnackbarHiding}
    label={this.state.EmailVerificationSnackbarLabel}
    timeout={10000}
    onTimeout={this.handleSnackbarHiding}
  />)
: undefined}

{this.state.showEmailVerificationSentMessage
  ? (<Snackbar className="sm-snackbar" action="Dismiss" type="cancel"
  active={this.state.showSnackbar === "sent"}
  onClick={this.handleSnackbarHiding}
  label="Verification email sent!"
  timeout={10000}
  onTimeout={this.handleSnackbarHiding} />)
  : undefined}
```

src/frames/Dashboard.tsx
chris-ramon: Following the comment above of avoiding `this.state.showSnackbar`, once the email verification is sent:
```typescript
remoteservice.defaultService().auth().currentUser.sendEmailVerification()
  .then(() => {
    this.setState({ ...this.state, showEmailVerificationSentMessage: true });
  });
```

src/index.tsx
```diff
@@ -74,11 +74,13 @@ Firebase.auth().onAuthStateChanged(function (user: Firebase.User) {
...
     const lastUser = state.session.user;
     const location = state.routing.locationBeforeTransitions;
-    const newLocation = {...location, ...{pathname: "/" }}; // Doing this will pass along any query parameters that may exist.
+    const newLocation = { ...location, ...{ pathname: "/" } }; // Doing this will pass along any query parameters that may exist.
     // If there is a user, set it
     if (user) {
         if (!lastUser || lastUser.userId !== user.uid) {
-            store.dispatch(setUser(new FirebaseUser(user)));
+            const dashboardUser = new FirebaseUser(user);
+            user.providerData[0].providerId !== "password" && (dashboardUser.emailVerified = true); // Prevents github users from getting the verify email notification
```
chris-ramon: This line should be removed ?, seems doing nothing.

src/frames/Dashboard.tsx
```diff
@@ -216,19 +247,29 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+  }
+
+  handleVerifyEmailClick() {
+    remoteservice.defaultService().auth().currentUser.sendEmailVerification();
+    this.setState({ ...this.state, showSnackbar: "sent" });
+  }
+
+  handleSnackbarHiding() {
+    this.setState({ ...this.state, showSnackbar: false });
+    this.props.setUser(this.props.user);
```
AnkRaiza: not sure about this maybe you should be using a state variable user instead of changing the props on compopnentupdate

src/frames/Dashboard.tsx
```diff
@@ -128,6 +151,15 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
     await this.props.getSources();
   }
 
+  componentDidUpdate(previousProps: DashboardProps, previousState: DashboardState) {
+    if (!previousProps.user.emailVerified && !previousState.showSnackbar) {
+      this.props.user.emailVerified = true;
```
AnkRaiza: avoid changing component's props like this, props should be inmutable and pass from parent to children, here maybe we can use a dispatch(setUser({...user, emailVerified: true})) or this.props.setUser({...user, emailVerified: true}).

src/frames/Dashboard.tsx
dmarvp: done.

src/frames/Dashboard.tsx
dmarvp: I have renamed the constant to emailVerificationSnackbarLabel, but I can't change it into a static variable, as it's referencing `this.handleVerifyEmailClick`.

src/frames/Dashboard.tsx
dmarvp: I see a couple of issues with this approach:
1. Conditional rendering of Snackbars will re-render the component based on the state. Whenever this.state.showEmailVerificationSentMessage changes, Snackbars won't disappear using the animation controlled by their "active" property, but they would appear/disappear abruptly instead.
2. I think this approach decreases readability, since we are conditionally rendering two snackbars using two variables instead of rendering the two and showing them based on a single variable. Also, we will always have to add one state variable be it showEmailVerificationSentMessage or showSnackbar.

src/frames/Dashboard.tsx
dmarvp: changed it to use the "then" callback. Haven't changed the state variable name, comments below.

src/index.tsx
dmarvp: The issue stated that only users logging in with email and password should receive the verification emails. These lines are preventing other providers (github or any other in the future) to receive the notification saying that their email hasn't been verified by changing their emailVerified property to true before putting it in the store.

src/frames/Dashboard.tsx
chris-ramon: right!, is using `this.handleVerifyEmailClick`, missed it, better to not have it as property state, because it does not change, therefore wrap it inside a function, like:
```typescript
emailVerificationSnackbarLabel() {
  return (<div>
    Your email is not yet verified - please click on the link in the email we sent to you at signup. Or click
    <a onClick={this.handleVerifyEmailClick} className="cursor-pointer"> here </a>
    to receive another verification email
    </div>);
}
```

src/frames/Dashboard.tsx
chris-ramon: good call on using the `active` property instead, but I still think is simpler to use existing variable: `this.props.user.emailVerified` and new state boolean variable: `this.state.showEmailVerificationSentMessage`, instead of having a new state variable with multiple sub-states, therefore something like this might work:

```typescript
<Snackbar className="sm-snackbar" action="Dismiss" type="cancel"
    active={!this.props.user.emailVerified && !this.state.showEmailVerificationSentMessage}
    onClick={this.handleSnackbarHiding}
    label={this.state.EmailVerificationSnackbarLabel}
    timeout={10000}
    onTimeout={this.handleSnackbarHiding}/>

<Snackbar className="sm-snackbar" action="Dismiss" type="cancel"
  active={this.state.showEmailVerificationSentMessage}
  onClick={this.handleSnackbarHiding}
  label="Verification email sent!"
  timeout={10000}
  onTimeout={this.handleSnackbarHiding} />
```

src/index.tsx
chris-ramon: ok cool, I think it reads better:

```typescript
// Email is verified for authenticated user via providers different than `email/password`. Eg. GitHub
if (user.providerData && user.providerData.length && user.providerData[0].providerId !== "password") {
  dashboardUser.emailVerified = true
}

Comment (dmarvp) [reply src/index.tsx]: done

File: src/frames/Dashboard.tsx [hunk 8]
```diff
@@ -216,19 +245,39 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+  emailVerificationSnackbarLabel(): any {
+    return (<div>
+      Your email is not yet verified - please click on the link in the email we sent to you at signup. Or click
+      <a onClick={this.handleVerifyEmailClick} className="cursor-pointer"> here </a>
+      to receive another verification email
+      </div>);
+  }
+
+  async handleVerifyEmailClick() {
+    remoteservice.defaultService().auth().currentUser.sendEmailVerification()
```
jperata: await remoteservice.defaultService().auth().currentUser.sendEmailVerification();
this.setState({ ...this.state, emailVerificationStatus: "sent" });

src/frames/Dashboard.tsx
```diff
@@ -85,15 +91,22 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+    this.handleVerifyEmailClick = this.handleVerifyEmailClick.bind(this);
+    this.handleSnackbarHiding = this.handleSnackbarHiding.bind(this);
+    this.emailVerificationSnackbarLabel = this.emailVerificationSnackbarLabel.bind(this);
+
+    this.state = {
+      showModal: false,
+      emailVerificationStatus: "loading"
+    };
+
+    console.log("state", this.state.emailVerificationStatus);
```
jperata: Remove log

src/frames/Dashboard.tsx
dmarvp: done

src/frames/Dashboard.tsx
dmarvp: Tried this approach but fell into the issue described in #505 and multiple renders of the snackbar. Until that one is solved, we are using a flag outside of the component and a state variable that shows the flow of the snackbars: loading | asking | sent | done.

src/frames/Dashboard.tsx
dmarvp: changed it so that we are no longer modifying the props inside the component.

src/frames/Dashboard.tsx
dmarvp: done

src/frames/Dashboard.tsx
dmarvp: done

chris-ramon:
LGTM :+1: :ship:

PR #494, opened 2017-11-22
474 email

dmarvp:
Relates #474 

Changes in this pull request include:

- The toast component now uses redux
- The logic to show a toast was moved to LoginPage.ts

PR #507, opened 2017-12-04
Fixes, toast, intercom

### LINKED ISSUE #496: Add intercom chat widget
Please include javascript snippet to show intercom chat on all Dashboard pages (except login page), the widget needs to include Bespoken app ID and other parameters I will submit internally by Slack. Thanks.

AnkRaiza:
Fixes #496 

- minor fixes for dashboard
- add intercom chat widget
- add toast component (using redux implementation)


chris-ramon:
- The main motivation for adding a new Toast component was to be used on https://github.com/bespoken/dashboard/pull/494 @AnkRaiza ? 
- Since we are using `react-toolbox` which comes with `Snackbar` component, seems good enough for our current use cases.
- Also we have a [wrap component](https://github.com/bespoken/dashboard/blob/master/src/components/Snackbar.tsx) around `Snackbar` which I would not recommend to use and perhaps plan to remove it since it adds unnecessary extra-layer.
- Also is there are motivation for getting together intercom changes within this PR ?

src/frames/Dashboard.tsx
```diff
@@ -234,6 +235,17 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
     return (
       <Layout header={true}>
+        {
+            globalWindow && globalWindow.Intercom("boot", {
+                app_id: "ah6uagcl",
+                name: this.props.user.displayName,
+                email: this.props.user.email,
+                signUpDate: this.props.user.userId,
+                skillsAmmount: this.props.sources.length,
+                usingMonitoring: this.props.sources.some(source => source.monitoring_enabled),
```
chris-ramon: I would rename `usingMonitoring` -> `monitoringEnabled` and `usingValidation` ->  `validationEnabled` for clarity.

AnkRaiza: as per Ivan's requirement this are the names on intercom's board so i had to use these names

AnkRaiza:
The toast component was for a later use as well, but also we wanted to use it on the #494, i agree that we have unnecessary layer around snack i think we can create an issue in order to remove it (the extra layer), the intercom widget was included in this PR because it was a 2 liner and this PR was created to add just the Toast, minor fixes and intercom widget

chris-ramon:
lgtm :+1:

PR #517, opened 2017-12-22
Landing topbar

AnkRaiza:
Relates #522 

- Header
- Header Test
- Package.json


jperata:
Add screenshots

AnkRaiza:
this change is not a UI change but a behavior change, so i screenshots would look the same as before, when checking just make sure the utton for the audio or log tab wont show if you have no logs or audio metrics. the ui changes will be made on the landing top bar UI ticket

Issue #518, opened 2017-12-28
login with amazon dashboard backend changes


dmarvp:
1.  On login page:
- [x] on click "login with amazon" button should open the login with amazon popup
- [x] once a user is authorized, we should exchange the access token for the user amazon profile
- [x] call the authToken endpoint from source-api passing the amazon user information along
- [x] when coming back, the query should include a custom authentication token, use it to authenticate with firebase using the auth.signInWithCustomToken function
- [x] open dashboard

2. On validation page:
- [x] when clicking on get a new token, should redirect to the silent echo "link_account" and specify a different "callback_url" from source-api (linkAVS) and the dashboard user id.
- [x] make sure to pass the silent echo user id (if it exists), so that a new one is not created all the time

3. [x] add/modify tests accordingly

Login with amazon workflow:

![login with amazon v4](https://user-images.githubusercontent.com/6411740/34833695-1099c6e6-f6be-11e7-8e08-3c820314f122.jpg)

Link account workflow:

![link with amazon](https://user-images.githubusercontent.com/6411740/34419377-90997c30-ebd1-11e7-9071-491b8f2d5b3d.jpg)

dmarvp:
Updated the ticket to reflect the latest workflow. 
Estimated time 1.5 days.

dmarvp:
To test it:

- With an amazon user:
1. Go to: https://dev-dashboard-bbcee.firebaseapp.com/dashboard/login
2. Click on login with amazon button
3. Enter your amazon credentials
4. Verify that the new amazon landing page appears

![loginwithamazon](https://user-images.githubusercontent.com/6411740/35125197-b52857d2-fc76-11e7-83dd-29fa2899263f.gif)

- Without an amazon user:
1. Go to: https://dev-dashboard-bbcee.firebaseapp.com/dashboard/login
2. Enter your credentials and Click on login OR login with github
3. Verify that the previous landing page appears

![normallogin](https://user-images.githubusercontent.com/6411740/35125296-1c007dc2-fc77-11e7-9e04-55f82d0cfd64.gif)

PR #543, opened 2018-01-16
Login with Amazon

dmarvp:
relates #518 
- added the sign in with custom token firebase method
- added the calls to virtual device and source api related to the amazon login
- when logging in, an access token is used to get the amazon profile
- we call source-api with that profile to get an auth token
- if no user is present, one is created
- after logging in with the auth token, a dispatch to setAmazonFlow is made


harness/index.html
```diff
@@ -69,6 +69,19 @@
 <script src="https://unpkg.com/react-dom@15.3.2/dist/react-dom.js"></script>
 <!-- Import the dashboard dependency -->
 <script src="/scripts/dashboard.js"></script>
+<!-- Login with Amazon -->
+<div id="amazon-root"></div>
+<script>
+    window.onAmazonLoginReady = function() {
+    amazon.Login.setClientId('amzn1.application-oa2-client.675b266c8bca4d408e5dceed285e18bd');
```
jperata: use an env vairable

src/actions/session.ts
```diff
@@ -16,13 +16,25 @@ export type SetUser = {
   user: User | undefined
 };
 
+export type SetAmazonFlow = {
```
jperata: AmazonFlowFlag to differentiate from the fucntion

src/pages/LoginPage.tsx
```diff
@@ -93,6 +97,9 @@ export class LoginPage extends React.Component<LoginPageProps, LoginPageState> {
 
     handleFormLoginWithAmazon() {
         this.props.loginWithAmazon()
+            .then((user) => {
```
jperata: try {        
  await this.props.loginWithAmazon();
  this.props.setAmazonFlow(true); 
} catch( err: Error) {
                 this.state.error = err.message;
                 this.setState(this.state);
}

src/services/auth.ts
```diff
@@ -19,8 +20,31 @@ namespace auth {
...
+        const accessToken = await amazonAuthorize(options);
+        const amazonProfile: any = await amazonRetrieveProfile(accessToken);
+        const authResponse = await source.getAuthToken(amazonProfile.PrimaryEmail, amazonProfile.Name);
+        return loginWithCustomToken(authResponse.authToken);
+    }
+
+    function amazonAuthorize(options: any): Promise<string> {
+        return new Promise(function (resolve, reject) {
+            globalWindow.amazon.Login.authorize(options, function (result: any) {
+                if (result.error) { reject(result.error); };
```
jperata: while reject here ensure state of the promise is always reject, it doesn't actually end the code flow, 
```
 if (result.error) {
 reject(result.error); 
 return;
}; 
```
will have the same effect but any addition after the return is ensured to not be run

src/services/auth.ts
```diff
@@ -19,8 +20,31 @@ namespace auth {
...
+                resolve(result.access_token);
+            });
+        });
+    }
+
+    function amazonRetrieveProfile(accessToken: string) {
+        return new Promise(function (resolve, reject) {
+            globalWindow.amazon.Login.retrieveProfile(accessToken, function (result: any) {
+                if (result.error) { reject(result.error); };
```
jperata: same

src/services/auth.ts
```diff
@@ -135,6 +159,20 @@ namespace auth {
             });
     }
 
+    export function loginWithCustomToken(token: string, auth: remoteservice.auth.Auth = remoteservice.defaultService().auth(), localStorage: LocalStorage = new BrowserStorage()): Promise<User> {
+        return auth.signInWithCustomToken(token)
```
jperata:
```
const user: remoteservice.user.User = await auth.signInWithCustomToken(token);
ReactGA.event({
   category: "Authorization",
   action: "Login With Custom Token"
});
let modelUser: User = new FirebaseUser(user);
identify(modelUser, "custom token");
ocalStorage.setItem("user", JSON.stringify(modelUser));
return modelUser;
```

src/services/auth.ts
```diff
@@ -135,6 +159,20 @@ namespace auth {
...
+            .then(function (user: remoteservice.user.User) {
+                ReactGA.event({
+                    category: "Authorization",
+                    action: "Login With Custom Token"
+                });
+                let modelUser: User = new FirebaseUser(user);
```
jperata: const if identify doesn't change modelUser

src/services/source.ts
```diff
@@ -73,6 +80,33 @@ export namespace source {
...
     }
 
+    /**
+     * This service will return an authorization token for a given email.  If the user does not exist,
+     * it will create it first.
+     *
+     * @param email the user's email.
+     * @param displayName the user's display name (optional)
+     */
+    export function getAuthToken(email: string, displayName: string): Promise<AuthResult> {
```
jperata: shouldnt this be diplayName?: string
according to your comment above

src/services/source.ts
```diff
@@ -73,6 +80,33 @@ export namespace source {
...
+        const query: Query = new Query();
+        query.add({parameter: "email", value: email});
+        query.add({parameter: "displayName", value: displayName});
+
+        return fetch(AUTH_TOKEN_URL, {
```
jperata: const result = await fetch ...

harness/index.html
dmarvp: Can't use an env variable in plain html. We should be safe, as the LWA uses both the client id and a whitelist to get the urls that can use that client id.

src/services/auth.ts
```diff
@@ -4,7 +4,7 @@ import { FirebaseUser } from "../models/user";
 import { User, UserDetails } from "../models/user";
 import { BrowserStorage, LocalStorage } from "../store/local-storage";
 import browser from "../utils/browser";
-import { remoteservice } from "./remote-service";
+import remoteService, { remoteservice } from "./remote-service";
```
jperata: remoteService (S capital) is just an alias, using remoteservice everywhere should work

Issue #548, opened 2018-01-19
Link AVS to user from the amazon flow


dmarvp:
Modify the landing page UI to include the validation token, it has to go before the vendor id.
Disable the "get started" button until both the validation token and vendor id have been added.
Call virtual-device and get a validation token.

dmarvp:
To try this, log in into dashboard using the amazon flow. Then click on the link to get a token and logging again if necessary or give your permissions on the amazon developer page. Click on the link to get the vendor Id and paste it on the corresponding textbox. See how the button enables just then.

![validation token](https://user-images.githubusercontent.com/6411740/35454923-00e829cc-029e-11e8-9c40-1052f142e5db.gif)

dmarvp:
@jperata moving this to Done - QA needed, please take a look when you can.

PR #550, opened 2018-01-22
Link AVS from Amazon Pane

dmarvp:
relates #548 

- Added a subscription to save any changes done to the "session" state of the store (this prevents losing the amazonFlow on refresh)
- Adds the validation token text input
- Links to virtual device to get the validation token
- The get started button only enables when both the validation token and vendor id are submitted
- Modified tests and themes accordingly


src/components/AmazonVendorPane.tsx
```diff
@@ -36,29 +65,52 @@ export default class AmazonVendorPane extends React.Component<any, any> {
...
+                                Validation Token
+                                <TooltipButton className={VendorPaneStyle.info_button} icon={"info"} tooltip={"In order to get the most out of our validation features, you will be redirected to the amazon developer page and asked to give us access to your skills information."} theme={TooltipTheme}/>
+                            </h4>
+                            <span>
+                                <Input
+                                    className={classNames(InputTheme.validation_token_input)}
+                                    style={{ color: "#000" }}
+                                    theme={InputTheme}
+                                    label="Validation token"
+                                    value={this.state.token}
```
jperata: this will render twice, first without the token, and then when the promise in component did mount resolves will render again once the state is setup. A loading while this is false may be needed

dmarvp: It doesn't, it just updates from a blank input to a filled one as soon as it gets the data.

PR #552, opened 2018-01-22
New Loader component

dmarvp:
#relates 551

- Adds a component to show a loader div that covers the whole page
- Added the component to the login page (when logging in with amazon)
- The component can be added to the dashboard component to have a globally available loading div if needed.
- In order to make it work with the amazon login, the process was divided in two parts: the authorizing process, and the login into dashboard itself. The loader is shown right after the amazon authorization.
- Added / modified tests accordingly.

lwa


src/pages/LoginPage.tsx
```diff
@@ -79,38 +80,42 @@ export class LoginPage extends React.Component<LoginPageProps, LoginPageState> {
         // Show some feedback in the link
     }
 
-    handleFormSubmit(email: string, pass: string) {
-        this.props.login(email, pass)
-            .catch((err: Error) => {
-                this.state.error = err.message;
-                this.setState(this.state);
-            });
+    async handleFormSubmit(email: string, pass: string) {
```
jperata: :+1:

dmarvp:
- Adds a component to show a loader div that covers the whole page
- Added the component to the login page (when logging in with amazon)
- The component can be added to the dashboard component to have a globally available loading div if needed.
- In order to make it work with the amazon login, the process was divided into two parts: the authorizing process, and the login into dashboard itself. The loader is shown right after the amazon authorization.
- Added / modified tests accordingly.

Issue #532, opened 2018-01-03
Validation page - general UI


AnkRaiza:
![image9](https://user-images.githubusercontent.com/11048365/35590440-e3fd5cbc-05d4-11e8-86c0-0272addfe50e.jpg)

in order to access this page, login with amazon and select a source, if there are no sources created, create a new source and then select it from the source list page. (the amazon flow is soon to be merge, so the alternate flow will be: login with amazon and then you will be redirected to validation page if you have at least 1 source on amazon)

this page has all the validation components on it

- validation parent component (monitoring button, title, un skill button)
- validation test component (input and expected titles and inputs and add row button)
- validation result component (actual title and result rows)

also in this page the parent backend and function for validation are working so we have here tickets #527, #528, #529, #530 resolved

robertonotte:
Re-tried today and I get to to this screen.
Still, I can't test it.
![validation-screen](https://user-images.githubusercontent.com/34077077/35524510-c8f2c456-04ef-11e8-8705-1c36b157b05a.jpg)

AnkRaiza:
Hi Roberto remember that here you are testing the UI, you can create a source and go to the validation page by clicking on it, the tickets to complete the amazon flow are about to be merge but in the meantime you can get into the validation page by creating a source and clicking on it on the source list page (wich is the one you see)

robertonotte:
![validation-changes](https://user-images.githubusercontent.com/34077077/35997332-29f565ac-0ce7-11e8-9552-d5a016fbe876.png)

PR #565, opened 2018-02-09
Complements

AnkRaiza:
Relates #532


src/components/AmazonVendorPane.tsx
```diff
@@ -197,7 +212,7 @@ export default class AmazonVendorPane extends React.Component<AmazonVendorPanePr
                         </a>
                     )}
                 </LandingAmazonPageTwoPane>
-                {this.state.loading && <Loader />}
+                {(this.state.loading || !this.props.finishLoading) && <Loader />}
```
jperata: why do we need two different states instead of just setting this.state.loading?

AnkRaiza: yes, loading is the component state, finishloading is the parent component loading status (sources loading, change of source etc.)

jperata: rename to parentFinishedLoading

PR #585, opened 2018-02-15
Fixes a bug when skipping with an empty vendor Id

dmarvp:
relates #558 

- Does not save anything to the database when clicking skip.


src/components/AmazonVendorPane.tsx
```diff
@@ -98,6 +99,16 @@ export default class AmazonVendorPane extends React.Component<AmazonVendorPanePr
         this.setState({ ...this.state, vendorID: value });
     }
 
+    async handleSkip() {
+        this.setState((prevState) => {
+            return { ...prevState, loading: true };
+        });
+        this.props.setAmazonFlow(false);
```
jperata: either this is async which means this needs an await here, or is not async which means handleSkip is not async.

PR #593, opened 2018-02-24
right panel and source selector changes

AnkRaiza:
image


src/components/Header.tsx
```diff
@@ -94,6 +96,8 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
   }
 
   render() {
+    const currentLocation = globalWindow.location;
+    const isSourceListPage = currentLocation.pathname === "/dashboard/skills" || currentLocation.pathname === "/dashboard/skills/";
```
dmarvp: same condition in the or statement?

src/components/RightPanel.tsx
```diff
@@ -10,30 +10,56 @@ export default class RightPanel extends React.Component<any, any> {
...
-                    <div>
-                        <p>We value your time: <strong>5 min set-up</strong></p>
-                        <Button
-                            className={VendorPaneStyle.get_started}
-                            theme={ButtonTheme}
-                            raised={true}
-                            primary={true}
-                            onClick={this.props.handleGetStarted}
-                            label="Get Started - FREE"/>
+        return this.props && this.props.type && this.props.type === "sourceListPage" ?
```
dmarvp: I would suggest having two different components instead of one. That way you could save the if condition and also would be good for maintainability.

src/components/SourceSelector/SourceSelectorItemComponent.tsx
```diff
@@ -58,13 +62,37 @@ export default class SourceSelectorItem extends React.Component<SourceSelectorIt
...
-            ...prevState,
-            isSourceUp,
-        }));
+        try {
+            const sourceId = this.props.source && this.props.source.id;
+            const result = sourceId && this.props.source.monitoring_enabled && await MonitoringService.getSourceStatus(sourceId);
+            const isSourceUp = result && result.status === "up" ? true : false;
+            const query: LogQuery = new LogQuery({
+                source: this.props.source,
+                startTime: moment().subtract(7, "days"), // change 7 for the right time span once implemented
```
dmarvp: add TODO

src/components/Header.tsx
AnkRaiza: the second one has an extra slash "/" at the end, since both are the same route

src/components/RightPanel.tsx
AnkRaiza: i'll add a TODO for this one we don't wanna use extra time right now to maintain a panel that will go away when we get to work on the actual graphs

src/components/Header.tsx
dmarvp: then use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/components/RightPanel.tsx
dmarvp: ok

src/components/Header.tsx
AnkRaiza: if i  use that it would be right for "/dashboard/skills/skill-id" and that would be incorrect

dmarvp: regex is your friend: /^\/dashboard\/skills\/?$/.test(currentLocation.pathname)

PR #616, opened 2018-03-10
right panel integrated plus issues fix

AnkRaiza:
creates the new right panel integrated view on dev, show message and link on prod, fixes validation page to enable monitoring without token, fixes tabs hidden behaviour

image


src/components/SourceSelector/SourceSelectorGlobalStats.tsx
```diff
@@ -0,0 +1,138 @@
...
+            query.add(new FillGapsParameter(true));
+            const timeSummary = await LogService.getTimeSummary(query);
+            const responseTimeService: any = await LogService.getResponseTimeSummary(query);
+            const dailyEventsAverage = timeSummary.buckets.reduce((acum: LogService.TimeSummary, bucket: any, index, array) => {
+                if (index === array.length - 1) {
+                    return (acum + bucket.count) / array.length;
+                }
+                return acum + bucket.count;
+            }, 0);
+            const responseTimeAverage = responseTimeService.data.reduce((acum: any, bucket: any, index: any, array: any) => {
```
jperata: all this acum can be simple by doing 
return acum +  (bucket.avgReponseTime/array.length)

src/components/SourceSelector/SourceSelectorGlobalStats.tsx
```diff
@@ -0,0 +1,138 @@
...
+            const { source, startDate, endDate } = this.props;
+            const query: Query = new Query();
+            query.add(new SourceParameter(source));
+            query.add(new StartTimeParameter(startDate));
+            query.add(new EndTimeParameter(endDate));
+            query.add(new GranularityParameter("hour"));
+            query.add(new FillGapsParameter(true));
+            const timeSummary = await LogService.getTimeSummary(query);
+            const responseTimeService: any = await LogService.getResponseTimeSummary(query);
+            const dailyEventsAverage = timeSummary.buckets.reduce((acum: LogService.TimeSummary, bucket: any, index, array) => {
```
jperata: same

src/components/SourceSelector/SourceSelectorGlobalStats.tsx
```diff
@@ -0,0 +1,138 @@
...
+                if (index === array.length - 1) {
+                    return (acum + bucket.count) / array.length;
+                }
+                return acum + bucket.count;
+            }, 0);
+            const responseTimeAverage = responseTimeService.data.reduce((acum: any, bucket: any, index: any, array: any) => {
```
jperata: same

src/components/SourceSelector/SourceSelectorGlobalStats.tsx
jperata: then you will  need to replace array for the actual variable, so that you only need acum and bucket as parameters

PR #638, opened 2018-03-27
1 2 step login

AnkRaiza:
New 1 2 steps login
validation page right panel


src/components/AmazonVendorPane.tsx
```diff
@@ -53,87 +40,54 @@ export default class AmazonVendorPane extends React.Component<AmazonVendorPanePr
...
-            const stepsCompleted = this.state.token && this.state.vendorID;
-            if (stepsCompleted) {
+        let userDetails: UserDetails = await auth.currentUserDetails();
+        // Only the first time the user clicks the integration we will try to create his sources from amazon.
+        if (userDetails && userDetails.smAPIAccessToken && !userDetails.vendorID) {
+            this.props.setLoading(true);
+            const vendorsId = await SourceService.getVendorIds(userDetails.smAPIAccessToken);
+            // for now we default to the first vendorID, later we'll add a way for the user to pick vendor id
+            const vendorID = vendorsId && vendorsId.length && vendorsId[0].id;
+            await auth.updateCurrentUser({ vendorID });
```
jperata: This will set the vendorID to false if the validation fail, if it does fail it should end the flow right there.

src/components/SourceSelector/SourceSelectorParentComponent.tsx
```diff
@@ -57,24 +72,30 @@ export class SourceSelector extends React.Component<SourceSelectorProps, SourceS
...
+            userDetails: undefined,
         };
 
         this.handleSourceClick = this.handleSourceClick.bind(this);
         this.handleDeleteSource = this.handleDeleteSource.bind(this);
         this.handleLoadingChange = this.handleLoadingChange.bind(this);
     }
 
-    componentDidUpdate (prevProps: SourceSelectorProps) {
+    async componentDidUpdate (prevProps: SourceSelectorProps, prevState: SourceSelectorState) {
```
jperata: no awaits in this method

src/components/AmazonVendorPane.tsx
AnkRaiza: if the validation fail it will set it to undefined, which is the current value so it should not be an issue, but i think it's more readable ending it if it fails i'll change it

Issue #645, opened 2018-04-04
Refresh token/get new token


AnkRaiza:
add a way to refresh / create a new token for validation

PR #647, opened 2018-04-05
add refresh token and prevent enable validation on unverified emails

AnkRaiza:
relates #639 #645 
untitled


AnkRaiza:
@jkelvie sorry i closed the other one accidentally this is the same with screenshot attached

PR #670, opened 2018-04-13
fix wrong monitor flag

AnkRaiza:


src/components/Validation/ValidationComponentGlobalStats.tsx
```diff
@@ -36,6 +37,15 @@ export default class ValidationPageGlobalStats extends React.Component<Validatio
         this.props.goTo("/skills/" + this.props.source.id + "/logs");
     }
 
+    shouldComponentUpdate (nextProps: ValidationPageGlobalStatsProps) {
+        if (SourceUtils.equals(nextProps.source, this.props.source) &&
```
jperata: just return the negated condition
!(SourceUtils.equals(nextProps.source, this.props.source) &&
nextProps.startDate.diff(this.props.startDate, "minutes", true) <= 1
&& nextProps.endDate.diff(this.props.endDate, "minutes", true) <= 1)

PR #689, opened 2018-04-27
dropdown height, source rename, catch error on source api

AnkRaiza:


src/components/AmazonVendorPane.tsx
```diff
@@ -51,7 +51,10 @@ export default class AmazonVendorPane extends React.Component<AmazonVendorPanePr
         // Only the first time the user clicks the integration we will try to create his sources from amazon.
         if (userDetails && userDetails.smAPIAccessToken && !userDetails.vendorID) {
             this.props.setLoading(true);
-            const vendorsId = await SourceService.getVendorIds(userDetails.smAPIAccessToken);
+            const vendorsId = await SourceService.getVendorIds(userDetails.smAPIAccessToken).catch(err => {
+                // if there is something wrong getting the vendor ids (token invalid/expired/not existent) just continue with no sources creation or user update
+                return [];
```
jperata: just return, componentDidMount results are not received by anyone

src/components/Title.tsx
```diff
@@ -3,55 +3,65 @@ import Source from "../models/source";
...
-        if (selectedSource) {
-            this.props.handleUpdateSource && await this.props.handleUpdateSource({...selectedSource, sourceType});
+    componentDidMount () {
+        this.setState(() => ({
+            sourceName: this.props.source && this.props.source.name || "",
+        }));
+    }
+
+    handleTypeChange = (sourceType: string) => {
+        if (this.props.source) {
```
jperata: duplicated code in handleLocalChange and handleSourceNameChange, extract to function

src/components/Validation/ValidationParentComponent.tsx
```diff
@@ -90,7 +90,10 @@ export class ValidationParentComponent extends React.Component<ValidationParentC
         const userDetails: UserDetails = await auth.currentUserDetails();
         if (userDetails && userDetails.smAPIAccessToken && !userDetails.vendorID) {
             this.props.setLoading(true);
-            const vendorsId = await SourceService.getVendorIds(userDetails.smAPIAccessToken);
+            const vendorsId = await SourceService.getVendorIds(userDetails.smAPIAccessToken).catch(err => {
+                // if there is something wrong getting the vendor ids (token invalid/expired/not existent) just continue with no sources creation or user update
+                return [];
```
jperata: just return

src/pages/validation/ValidationPage.test.tsx
```diff
@@ -37,7 +37,7 @@ describe("Validation Page", function () {
         afterEach(function() {
             currentUserDetailsStub.restore();
         });
-        it("Tests that validation token is set.", function () {
+        it.skip("Tests that validation token is set.", function () {
```
jperata: we still use validation token, why we skip this one?

src/services/source.ts
```diff
@@ -295,12 +298,14 @@ export namespace source {
...
             });
-            const vendorsArray = await result.json();
-            if (vendorsArray) {
-                return Promise.resolve(vendorsArray);
-            } else {
-                return Promise.reject(new Error(result.statusText));
+            if (result.status === 200) {
+                const vendorsArray = await result.json();
+                if (vendorsArray) {
+                    return Promise.resolve(vendorsArray);
```
jperata: since function is using async await already you should be able to just return here

src/services/source.ts
```diff
@@ -295,12 +298,14 @@ export namespace source {
...
+                }
+                return Promise.resolve([]);
```
jperata: same

src/services/source.ts
```diff
@@ -295,12 +298,14 @@ export namespace source {
             }).catch(err => {
                 return err;
```
jperata: This error would be registered as correct response instead of an exception in the other side, do a reject or throw if needed

src/components/AmazonVendorPane.tsx
AnkRaiza: this is the return for the getVendorIds which should return the vendorIds array, i return empty on error because after this call we check for vendorsId && vendorsId.length && vendorsId[0].id;

src/components/Validation/ValidationParentComponent.tsx
AnkRaiza: same as above this is the return for getVendorIds not the return for componentDidmount

src/pages/validation/ValidationPage.test.tsx
AnkRaiza: you are right i thought this was for the smapi token which we no longer show on the page 👍

PR #693, opened 2018-05-02
show errors on reset password button

AnkRaiza:


src/pages/LoginPage.tsx
```diff
@@ -87,9 +87,17 @@ export class LoginPage extends React.Component<LoginPageProps, LoginPageState> {
...
 
-    handleResetPassword(email: string) {
-        this.props.resetPassword(email);
-        // Show some feedback in the link
+    async handleResetPassword(email: string) {
+        try {
+            await this.props.resetPassword(email);
+            // Show some feedback in the link
+        } catch (err) {
+            if (!email) {
```
jperata: first check this and send the error if needed, and then call the resetPassword function, doesn't made sense to make it fail if we know it would if there's no email there.

PR #695, opened 2018-05-02
await for state to finish saving

### LINKED ISSUE #694: Rename of Skills
This behavior was reported by John.

When I try to change the name and save it in prod, the new name does not seem to “take” - interestingly, I can update it on the settings page, and there it does seem to work.

AnkRaiza:
Fixes #694


jkelvie:
Is it possible to add a unit test for this?

AnkRaiza:
yes, adding unit test

src/components/Title.tsx
```diff
@@ -49,10 +49,10 @@ export class Title extends React.Component<TitleProps, any> {
 
     handleSourceNameChange = async (value: any) => {
         const {sourceName} = value;
-        this.updateSourceWithProps({name: sourceName});
-        this.setState(() => ({
+        await this.setState(() => ({
```
jperata: this.setState is not a promise, this shouldn't work.

The correct form to do this on react is to use the componentDidUpdate or to use the callback argument of the setState function.

PR #696, opened 2018-05-02
webPage for payment added

daisyzarzo:
relates to https://github.com/bespoken/dashboard/issues/688
image


src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+
+const BillCardStyle = require("./BillCardStyle.scss");
+
+export default class BillCard extends React.Component<any, any> {
```
AnkRaiza: Instead of any lets define which props will be accepted on this component, that way we can't just force props on it

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+    constructor(props: any) {
+        super(props);
+    }
+
+    render() {
+        return (
+            <div className={BillCardStyle.container} style={{ color: this.props.letterThema, background: this.props.containterTheme }}>
```
AnkRaiza: why do we need container theme and also BillCardStyle.container

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+                <div >
+                    <img src={this.props.uriImage} />
```
AnkRaiza: this img doesnt have alt in case the img takes too long to load or doesnt load at all.

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+                </div>
+
+                <div className={BillCardStyle.containerFooter} style={{ background: this.props.footerTheme }} />
```
AnkRaiza: if this is for the color then it is better this.props.footerColor or this.props.footerBackgroundColor

src/components/BillCard/BillCard.tsx
jperata: change letterThema to letterTheme

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+                <div  >
+                    <p>Free </p>
+                    <p>1 </p>
+                </div>
+                <div className={BillCardStyle.containerFooter} style={{ background: this.props.footerTheme }} />
```
AnkRaiza: same

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+                <div style={{ background: this.props.containterTheme }}>
+                    <p>3 </p>
+                    <p>10, 000 000 </p>
+                    <p>2</p>
+                </div>
+                <div className={BillCardStyle.containerFooter} style={{ background: this.props.footerTheme }} >
```
AnkRaiza: same

src/components/BillCard/BillCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+                </div>
+
+                <div className={BillCardStyle.containerFooter} style={{ background: this.props.footerTheme }} />
+                <div  >
+                    <p>Free </p>
+                    <p>1 </p>
+                </div>
+                <div className={BillCardStyle.containerFooter} style={{ background: this.props.footerTheme }} />
+                <div style={{ background: this.props.containterTheme }}>
+                    <p>3 </p>
```
AnkRaiza: lets add props for all the contents of this card (this 3, 10 000 000, and so on) these should be handle as props or else the body od the card itself should be handle as a prop, otherwise you cant use this card as a component for diferent prices/plans

src/pages/ComponentsPage.tsx
```diff
@@ -1,64 +1,110 @@
 ﻿import * as React from "react";
-import BarsChart from "../components/Graphs/Bar/BarsChart";
-import RadialBarChart from "../components/Graphs/Radial/ValidationRadialChart";
-
-const centerStyle = {
-    margin: "auto",
-    width: "50%",
-    padding: "10px"
-};
+import BillCard from "../components/BillCard/BillCard";
```
AnkRaiza: here we should not delete older components the purpose of this page is to show all component we are creating and how they work, your component should be an addition to this pagen ot a replacement

PR #715, opened 2018-06-22
remove status from boolean evaluation

AnkRaiza:
this was causing the graph to load instead of showing the down color and graph

image


src/pages/sourcepage/SourceUptime.tsx
```diff
@@ -113,13 +113,14 @@ export class SourceUpTime extends LoadingComponent.Component<{ summary: UpTimeDa
 
     render() {
         const {data} = this.state;
+        console.log(data);
```
dmarvp: do we need this log?

AnkRaiza: no we dont

dmarvp: then remove :P

PR #742, opened 2018-07-18
fix components page

AnkRaiza:


src/index.tsx
```diff
@@ -157,6 +157,10 @@ const render = function () {
     ReactDOM.render((
         <Provider store={store}>
             <Router history={history} onUpdate={onUpdate}>
+                {
+                    process.env.NODE_ENV !== "production" &&
```
ecruzado: maybe store this in a variable and reuse it

Issue #758, opened 2018-07-30
Enable or disable payment based on an environment variable.


dmarvp:
related to #746 

While the payment feature is completed, we must be able to disable everything related to it in production. Do this by creating an ENV variable called ENABLE_PAYMENT (default = False). 

If false: 
- The payment bar should not display.
- Monitoring limits and any other limits based on user plans should not be enforced.

Please be sure to use this environment variable instead of just looking at if the deploy was made to prod or dev.

MKen-li:
Move to QA Rejected.

1. The label is disable but the validations per plan are still in place.
![validation pricing](https://user-images.githubusercontent.com/39920441/43612463-089fc9f0-9672-11e8-90b0-66dc0ed51c47.png)

2. The buttons and drop down section have to much vertical white space.

MKen-li:
Mark as verified,
Build = dev-92

PR #759, opened 2018-07-30
Enable or disable payment based on an environment variable.

daisyzarzo:
relates to https://github.com/bespoken/dashboard/issues/758


src/components/Header.tsx
```diff
@@ -421,11 +425,14 @@ export class PageSwap extends React.Component<PageSwapProps, PageSwapState> {
   }
 
   render() {
+    const paymentEnable = " paymentEnable ";
```
AnkRaiza: lets use an actual local class instead of a global class here

src/index.tsx
```diff
@@ -164,6 +164,7 @@ const removeSource: LeaveHook = function () {
 };
 
 const render = function () {
+    const renderPayment = process.env.CV1 ? process.env.ENABLE_PAYMENT : true;
```
AnkRaiza: what is process.env.CV1

webpack.config.js
```diff
@@ -27,7 +27,7 @@ var free_sln = process.env.FREE_SLN;
 
 var standard_lln = process.env.STANDARD_LLN;
 var free_lln = process.env.FREE_LLN;
-
+var enable_payment = process.env.ENABLE_PAYMENT;
```
AnkRaiza: shouldnt this be also in the build variables?

AnkRaiza: my bad dint notice it was on the build variables

PR #787, opened 2018-08-14
Payment

AnkRaiza:
image
image
image


src/components/Payment/PaymentComponent.tsx
```diff
@@ -0,0 +1,320 @@
...
+        users: "1",
+        price: "$ 79",
+    },
+    enterprise: {
+        sources: "custom",
+        logs: "unlimited",
+        users: "1",
+        price: "custom",
+    },
+};
```
ecruzado: this should be hardcoded?

AnkRaiza: we'll have this for now as a const (i'll move it to the const file) but it is beign discused if this will reamin as a const or maybe be data taken from a DB

webpack.config.js
```diff
@@ -81,7 +81,7 @@ var buildVariables = {
 
     STANDARD_LLN: JSON.stringify(standard_lln),
     FREE_LLN: JSON.stringify(free_lln),
-    ENABLE_PAYMENT: JSON.stringify(enable_payment),
+    ENABLE_PAYMENT: JSON.stringify("true"),
```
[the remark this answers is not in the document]
AnkRaiza: yup, thanks

PR #788, opened 2018-08-16
add payment component and step

AnkRaiza:


src/components/BillCard/Payment.tsx
```diff
@@ -132,49 +110,46 @@ class SplitFormC extends React.Component<any & SplitFormCProps, any> {
...
-                                <button className={PaymentStyle.botton}>{"Subscribe"}</button>
-                            </fieldset>
-                        )
-                    }
+                    <fieldset >
+                        <label>
+                            <div className={PaymentStyle.card_container}>
+                                <CardElement style={{
+                                    base: {
+                                        // iconColor: "#666EE8",
```
ecruzado: remove it

src/components/BillCard/Payment.tsx
```diff
@@ -91,23 +69,23 @@ class SplitFormC extends React.Component<any & SplitFormCProps, any> {
...
         if (helper === "operation success") {
             if (this.props.subscriptionUpdate && this.props.sources) {
                 subscriptionPlans.checkMonitoredSkillLimits(this.props.planId, this.props.sources, user.email);
             }
-
-            setTimeout(() => {
-                this.redirect();
-            }, 3500);
+            this.props.handleSubscribe();
         }
```
ecruzado: It doesn't have the else statement

AnkRaiza: Yes we only switch step if the operation succeeded otherwise we use the message variable to show the error(s)

PR #801, opened 2018-08-22
fix refresh and add button to step contact

AnkRaiza:


src/components/Payment/PaymentComponent.tsx
```diff
@@ -291,6 +291,9 @@ export default class PaymentComponent extends React.Component<PaymentComponentPr
                 );
             }
             case stepEnum.payment: {
+                if (this.state.userSelectedPlan === this.state.selectedPlan) {
+                    this.handleCloseModal({});
```
ecruzado: you should probably validate the function's args rather than using an empty object here

PR #838, opened 2018-09-05
Env var

AnkRaiza:


scripts/prod-deploy.sh
```diff
@@ -15,8 +15,9 @@ if [[ "$ENV" == "dev" ]]; then
 fi
 if [[ "$ENV" == "staging" ]]; then
     echo "[CD] build staging started"
+    npm run build
```
jperata: either remove this and readdd it to the travis steps, or add it on the 3 deployments

Issue #831, opened 2018-09-04
Log limit email is not being sent


MKen-li:
**Preconditions**
Log in to the system

**Steps To Reproduce**
1. Select an skill and complete the integration process
2. Generate logs until you pass the limit per plan
3. Interact with dashboard 

**Expected behavior**
The warning email is sent

**Actual behavior**
The email is not sent

**Screenshots**
None.

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-120

**Additional context**

jperata:
Seems like it's only counting the logs at the moment if validation is enabled.

jperata:
Not working for 6 logs for a free user 
juan+payment@bespoken.io 
sourceid: de21da37-ac39-419c-babc-8ac58931ac61

jperata:
Seems like the change to get the benefits from the database is part of the issue, was at the time of the test at 1 million, reduced it to 4 , but it still failed.

jperata:
Marked as verified
Dashboard build: dev-128

Issue #827, opened 2018-09-03
Outage notifications have incorrect instructions


jkelvie:
**Steps To Reproduce**
1. Go to the skill -> settings page
2. Look at the text

**Expected behavior**
The text should say:
```
Enabling “Outage Notification” will:
* Ping the service at a five-minute interval (by default with an empty JSON payload)
* Notify the registered email address if the service does not respond
* Enable custom JSON to tailor the JSON payload
```
**Actual behavior**
The actual text is different, and incorrectly states a 1-minute interval.

It also has grammatical errors that should be fixed.

dmarvp:
Marked as verified.
Build: dev-127

Issue #825, opened 2018-09-03
The edit button area is shorter than the text box


MKen-li:
**Preconditions**
log in to the system.

**Steps To Reproduce**
1. Create a source with the max number of characters
2. Hover through the whole text

**Expected behavior**
The edit box and pen icon are show if the mouse is hover in all parts of the text box

**Actual behavior**
The edit button area is shorter than the text box

**Screenshots**
None.

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-144

**Additional context**

dmarvp:
Moving back to in Dev. Seems to be working correctly on Chrome and Safari, but the pencil icon does not appear in Firefox.

Also: can we add some space between the pen icon and the source name?

![sep-11-2018 22-32-59](https://user-images.githubusercontent.com/6411740/45400683-bb217a00-b612-11e8-9f50-b08d159452ca.gif)

dmarvp:
Verified in dev-129.

Issue #823, opened 2018-08-31
Technical change: Recover plan details from firebase.


dmarvp:
From: https://github.com/bespoken/dashboard/issues/793#issuecomment-416724794

We need to:

1. Get rid of the environment variables where we store the logs limitations: 
> process.env.STANDARD_LLN
> process.env.FREE_LLN
> process.env.PRO_LLN

2. Instead, use a new structure in firebase called "plans" inside firebase. E.g.:

```json
{
  "plans": [
    {
      "id": "free",
      "displayName": "Free",
      "benefits": {
        "monitoredSkills": 1,
        "logs": 1000000,
        "virtualDevices": 1
      }
    },
    {
      "id": "standard",
      "stripePlanId": <stripePlanId>,
      "displayName": "Standard",
      "benefits": {
        "monitoredSkills": 3,
        "logs": 10000000,
        "virtualDevices": 1
      }
    },
    {
      "id": "pro",
      "displayName": "Pro",
      "stripePlanId": <stripePlanId>,
      "benefits": {
        "monitoredSkills": 10,
        "logs": 50000000,
        "virtualDevices": 1
      }
    },
    {
      "id": "enterprise",
      "displayName": "Enterprise"
    },
  ]
}
```

3. for each user, add a new property "planId" (whenever the user changes plans) and benefits (for enterprise users **only**). E.g.:
```json
{
"user": {
  "id": <user_id>,
  "planId": <planId from the firebase plans collection>,
  "benefits": {
        "monitoredSkills": 50,
        "logs": 3000000,
        "virtualDevices": 15
      }
}
```

4. When login into the dashboard, we should:
    - Internally, retrieve the plan name and benefits from the plans collection (by planId)
        - Plan name is shown next to the user icon, and inside "My account".
    - If the user is an enterprise one, retrieve the benefits from it's own user object
    - Apply the limitations of those benefits inside the dashboard per user
    - If the user does not have the planId property, assume the user is in a free plan.

5. When opening the payment modal:
    - Plan names should come from the plans collection
    - The listed plan benefits should come from the benefits object of each plan

**NOTES**
- We won't be saving/using the ``stripeSubscribedPlanName`` property at the user level, since this would prevent us from changing the name of a plan globally. Instead, we will tie the user to a plan by a ``planId``.
- The other two properties from stripe on the user object (related to customer id and the stripe subscribed plan id) can remain there.
[...]

PR #851, opened 2018-09-10
use db for plan benefits, fix some ui issues

AnkRaiza:
#831 #823 #827 #757 #825 

change the payment module to use db for plans benefits and selections plus other minor bugs like text changes and style changes, i've separated those by commit here for easier review

Hover on long text
image

Outage notification
image


jperata:
Put a more descriptive title for the PR, remember once we squash this is the main message we will see

src/components/Payment/PaymentComponent.tsx
```diff
@@ -83,15 +84,31 @@ export default class PaymentComponent extends React.Component<PaymentComponentPr
 
     async componentDidMount() {
         const userDetails = await auth.currentUserDetails();
+        let { planId } = userDetails;
+        const plansObject = await getPlansData();
+        const plansData = this.formatPlansData(plansObject);
+        // leaving this here to support older users using the stripeSubscribedPlanName prop
```
jperata: we shouldn't have older users, since the property is not enabled in prod, if there is some issue with dev data from past builds we can edit those users (since is ourselves) or remove the user and create the account again as long is dev.

src/components/Payment/PaymentComponent.tsx
```diff
@@ -83,15 +84,31 @@ export default class PaymentComponent extends React.Component<PaymentComponentPr
...
+        if (!planId) {
+            planId = userDetails.stripeSubscribedPlanName;
+        }
+        this.setState(prevState => ({
```
jperata: each set state will generate a new render, seems unnecessary to do those along the way and not just to set all the state after you have awaited every api call, if you are doing it to avoid not rendering anything in case of errors, you should add some try's and catchs.

src/components/Payment/PaymentComponent.tsx
```diff
@@ -73,8 +74,8 @@ export default class PaymentComponent extends React.Component<PaymentComponentPr
             cardExpiration: "",
             cardCvc: "",
             focused: "",
-            standardPlanPrice: "",
-            proPlanPrice: "",
+            standardPlanPrice: "??",
```
jperata: why change this from empty to ??

src/components/Payment/PaymentComponent.tsx
```diff
@@ -255,47 +277,33 @@ export default class PaymentComponent extends React.Component<PaymentComponentPr
...
-                                    <RadioButton disabled={disableRadioEnterprise} theme={RadioButtonTheme} className={this.state.selectedPlan === "Enterprise" ? PaymentStyle.active : ""} label={
-                                        (
-                                            <PlanCard
-                                                userPlanId={this.state.selectedPlan}
-                                                planId={"Enterprise"}
-                                                uriImage={"https://bespoken.io/wp-content/uploads/2018/05/Enterprise.jpg"}
-                                                bottomStyle={true}
-                                                handleStepChange={this.handleStepChange}
-                                            />
+                                    {
```
jperata: Optional change , between tabs this looks too far to the left to be read easily, exporting to a variable could help with that.

src/components/Payment/PaymentComponent.tsx
AnkRaiza: external api call for price is not instant a instant render with plans information and another updating the price is just fine otherwise on cases with slow networks or slow external api call, the UX will not be great (waiting for popup to render)

src/components/Payment/PaymentComponent.tsx
AnkRaiza: the default price empty seems confusing we can use here 0 ?? loading or something else that suggest the prices are not loaded yet

src/services/subscription-plan.ts
```diff
@@ -1,49 +1,37 @@
...
-
-        switch (plan) {
-            case "Free": return F_LLN;
-            case "Standard": return S_LLN;
-            case "Pro": return P_LLN;
-            case "Enterprise": return Number.POSITIVE_INFINITY;
-            default: return 0;
-        }
+export const getLogsLimitNumber = async (plan: string) => {
+    let ref = remoteservice.defaultService().database().ref();
```
jperata: const

src/services/subscription-plan.ts
```diff
@@ -1,49 +1,37 @@
...
-            return F_SLN;
+export const getPlansData = async () => {
+    let ref = remoteservice.defaultService().database().ref();
+    const plansDataRef = await ref.child("/plans").once("value");
+    const plansData = plansDataRef.val();
+    return plansData || {};
+};
+
+export const getSourcesLimitNumber = async (plan: string) => {
+    let ref = remoteservice.defaultService().database().ref();
```
jperata: cosnt

src/services/subscription-plan.ts
```diff
@@ -1,49 +1,37 @@
...
-            ? process.env.FREE_SLN : 1;
-
-        if (plan === "Standard")
-            return S_SLN;
-        if (plan === "Pro")
-            return P_SLN;
-        if (plan === "Free")
-            return F_SLN;
+export const getPlansData = async () => {
+    let ref = remoteservice.defaultService().database().ref();
```
jperata: const

src/components/Payment/PaymentComponent.tsx
jperata: Can we preload this as part of the login process so that by that time is always loaded?

Issue #768, opened 2018-08-04
UI - Buttons update.


dmarvp:
Currently, we have buttons with different shapes, sizes, and content styles. Propose and implement an standard to share across the dashboard. The standard should include:

- Primary, secondary and tertiary buttons
- Variations for Default, Hover, Focused and Disabled states.

Some examples of what we currently have:

<img width="103" alt="image" src="https://user-images.githubusercontent.com/6411740/45574890-0fa53f00-b837-11e8-9ba1-423f546aa0b5.png">

<img width="282" alt="image" src="https://user-images.githubusercontent.com/6411740/45574923-25b2ff80-b837-11e8-9b3b-b5f156662e69.png">
<img width="130" alt="image" src="https://user-images.githubusercontent.com/6411740/45574931-2d72a400-b837-11e8-88cc-3e94ed8f37a4.png">
<img width="144" alt="image" src="https://user-images.githubusercontent.com/6411740/45574946-39f6fc80-b837-11e8-9c22-cc6b73d0abce.png">
<img width="339" alt="image" src="https://user-images.githubusercontent.com/6411740/45574956-4c713600-b837-11e8-9630-7041cf8106e6.png">
<img width="231" alt="image" src="https://user-images.githubusercontent.com/6411740/45922808-9d1e0a00-be9b-11e8-90e3-0479e557091c.png">

dmarvp:
@AnkRaiza Do we have styles for focused and hover as well? Also, the secondary button looks more prominent than the main one :)

dmarvp:
<img width="531" alt="image" src="https://user-images.githubusercontent.com/6411740/47446864-c6bf9f00-d781-11e8-8cee-5f1cc6661e5f.png">

buttons look ok but some changes are still needed:
- Settings button should have the same font.
- Under settings, "Advanced" button should be secondary.
- Cancel button should be a secondary button when:
    - deleting a source
    - changing password
- Buttons need a click animation like [this one](https://www.w3schools.com/howto/howto_css_animate_buttons.asp)

dmarvp:
Verified in dev-185.

PR #903, opened 2018-10-09
buttons ui

AnkRaiza:
#768 
image


dmarvp:
@AnkRaiza Do we have styles for focused and hover as well? Also, the secondary button looks more prominent than the main one :)

Issue #907, opened 2018-10-11
The "X" icon overlaps with the text when checking a wrong syntax


MKen-li:
**Preconditions**
Be on Yaml editor page.

**Steps To Reproduce**
1. type a bad syntax
2. click "check syntax" button

**Expected behavior**
The text and icon don't overlap with each other .

**Actual behavior**
The text and icon overlap with each other.

**Screenshots**
![check syntax error](https://user-images.githubusercontent.com/39920441/46820480-e8656300-cd4b-11e8-9084-c6d36f51c752.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-151

**Additional context**
None.

MKen-li:
Mark as verified
Build = dev-157

Issue #881, opened 2018-09-20
The hybrid source icon in validation page is not the correct one


MKen-li:
**Preconditions**
Log in to the system.

**Steps To Reproduce**
1. Create a hybrid source
2. Go to Validation Page

**Expected behavior**
The icon for hybrid source is the same as it is in source page.

**Actual behavior**
The icon doesn't have the correct size or positioning.

**Screenshots**
![hybrid source](https://user-images.githubusercontent.com/39920441/45832017-68a52500-bcc6-11e8-882b-e6ae24a678b2.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-138

**Additional context**

MKen-li:
Mark as verified
Build = dev-157

Issue #870, opened 2018-09-19
The Run button should be disable until the current test is completed


MKen-li:
**Preconditions**
Have a source created with a test.

**Steps To Reproduce**
1. Run the test

**Expected behavior**
The Run button is disable until the test is completed.

**Actual behavior**
The Run button is active when the test is running.

**Screenshots**
None

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-132

**Additional context**
None.

MKen-li:
Move back to dev,
Issue still happening

MKen-li:
Moving back to dev
After talking to @dmarvp we decided that the "check syntax button" should be disable as well when running the test

MKen-li:
Marked as verified
Build = dev-161

PR #910, opened 2018-10-15
fix validation ui and ux behabior

AnkRaiza:
#907 #881 #870


src/pages/validation/ValidationPage.tsx
```diff
@@ -324,7 +325,7 @@ export class ValidationPage extends React.Component<ValidationPageProps, Validat
                             self.setState({
                                 ...self.state,
                                 dialogActive: true,
-                                loadingValidationResults: false,
+                                loadingValidationResults: true,
```
jperata: ensure is this is correctly set on true or if this should also be related to "isTestLoading"

AnkRaiza: this is the first tick for the validation so it will always should have been true, the next values will be set on the pusher events or in the error event

Issue #955, opened 2018-10-31
Add validation message when no test is provided


MKen-li:
**Preconditions**
None.

**Steps To Reproduce**
1. Go to the Yaml editor
2. Delete the test section
3. Run the test

**Expected behavior**
The system shows the message "No test was found."

**Actual behavior**
no validation occurs an nothing happens

**Screenshots**
None.

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-187

**Additional context**

MKen-li:
Marked as verified
Build = dev-195

PR #974, opened 2018-11-14
"No test was found." message added to yaml

daisyzarzo:
relates to https://github.com/bespoken/dashboard/issues/955


daisyzarzo:
image
image

src/components/Validation/ValidationParentComponent.tsx
```diff
@@ -287,7 +287,12 @@ configuration:
         this.props.clearValidationResults();
         try {
             const parsedScript = await SourceService.parseYaml(this.props.yamlScript);
-            if (parsedScript) {
+            if (parsedScript && parsedScript._tests && parsedScript._tests.length === 0) {
```
jricaldi: just `!parsedScript._tests.length` instead `parsedScript._tests.length === 0`. Also apparently there is a missing "await" for `handleCheckSyntax` in line 187
@jperata why we are using `this.handleCheckSyntax.bind(this)`(bind thing) when handleCheckSyntax is an arrow function?

jperata: @jricaldi there are other functions on the same file that use the function notation, is an incomplete migration, we should consider not using the bind and move all to arrow functions, but it needs to be part of separate changes.

Issue #928, opened 2018-10-22
The GitHub login is not working


MKen-li:
**Preconditions**
None.

**Steps To Reproduce**
1. Go to the login page
2. Click on login with GitHub

**Expected behavior**
The GitHub login page is displayed and the user can log in correctly.

**Actual behavior**
The user can't log in and the system displayed the message "Error returned trying to log in."

**Screenshots**
![login github](https://user-images.githubusercontent.com/39920441/47304655-da85cc80-d5ec-11e8-9530-89224d42f513.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-162

**Additional context**
None.

dmarvp:
Apparently, it doesn't work in dev, but it does work in prod. Still, this is preventing QA from testing with a Github user in dev and should be fixed.

jricaldi:
@dmarvp the fix is already deployed but we have another could affect this.

MKen-li:
Marked as verified
Build = dev-199

PR #975, opened 2018-11-16
BUG: 928 The GitHub login is not working

jricaldi:
Relates to #928 

Changes:
--
- The User have already an email/password account, then try to login using github or amazon. Now, they are linked to the same user.
- The User have already and github or amazon account, then try to register a email/password. Now, they are linked to the same user.

### Important to know
It was necessary created a new Oauth App in Bespoken account, also change the GitHub configuration in firebase.

image


jricaldi:
working in the test src/services/auth.test.ts

Issue #755, opened 2018-07-27
The source page show the script had failed when it shouldn't


MKen-li:
**Preconditions**
Use a user that has already obtain a token

**Steps To Reproduce**
1. Create a source
2. Go to validation
3. Run a working script
4. Enable monitoring
5. Go back to Source Page

**Expected behavior**
Everytime the monitoring is enable show run and the result should reflect on the source page

**Actual behavior**
Monitoring takes about 30 to do the first run and in the meantime no result = bad result

**Screenshots**
![enable 1](https://user-images.githubusercontent.com/39920441/43348753-49cae8fc-91c1-11e8-94ca-4720f598d308.png)
![enable 2](https://user-images.githubusercontent.com/39920441/43348754-49e8d768-91c1-11e8-8d36-a7b5c56e17df.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-87

**Additional context**
None

MKen-li:
The waiting message is displayed but only after 5 seconds, before that it show that the test is good

MKen-li:
moving back to dev

MKen-li:
Marked as verified
Build = dev-204

PR #978, opened 2018-11-19
source page fixed, showing no failed script

daisyzarzo:
relates to  https://github.com/bespoken/dashboard/issues/755


src/services/source.ts
```diff
@@ -319,16 +320,22 @@ export namespace source {
...
+                        (err: Error): firebase.Promise<any> => {
+                            if (err) {
+                                reject(err);
+                            }
+                            return;
+                        });
+                if (source.validation_enabled !== undefined && !source.validation_enabled) {
+                    await MonitoringService.updateSourceStatus(source.id, "off");
+                }
+                resolve();
```
jperata: once remove the promise wrapper instead of resolve, return the source here

src/services/source.ts
```diff
@@ -319,16 +320,22 @@ export namespace source {
...
+            } catch (error) {
+                reject(error);
```
jperata: instead of reject you can remove the try catch since it will throw the error during await's, or print the error in console and then throw.

daisyzarzo:
image

Issue #962, opened 2018-11-07
Update the validation page to select from a list of virtual devices


dmarvp:
### Summary

As a user of dashboard with virtual devices
I want to be able to a virtual device 
so that I can run my validation scripts

### Acceptance Criteria

1. GIVEN: I'm a user with NO virtual devices
WHEN: I go to the validation page 
THEN: I still see the "Get validation token" link

2. GIVEN: 
    - I'm a user with NO virtual devices on the validation page. 
    - Alexa or Google is the platform chosen on the drop-down.
    - I have not reached my current limit of virtual devices

 WHEN: I click on "Get validation token"
 THEN: 
    - I'm taken to the login with Alexa or Login with Google page (depending on the platform chosen).
    - The flow continues as before but the virtual device is created under the new structure.

3. GIVEN: 
    - I'm a user with NO virtual devices on the validation page. 
    - "Hybrid Source" is the platform chosen on the drop-down. (Please change it to "Hybrid voice app")
    - I have not reached my current limit of virtual devices

 WHEN: I click on "Get validation token"
 THEN: 
    - My current script is saved.
    - I'm taken to the virtual devices manager page.

4. GIVEN:  I'm a user that has virtual devices.
 WHEN: I go into the validation page.
 THEN: 
    - I see a dropdown with the virtual devices for the chosen platform. Eg. If Alexa, I will only see alexa virtual devices. If Google, I will only see Google virtual devices. If both, I will see both types. 
    - If there's no virtual device for the chosen platform, I see the link "Get validation token"
    - The label shown is the virtual device name. The value is the virtual device token.
    - The copy button next to it will remain (will copy the virtual device token)
    - The refresh button tooltip says: Refresh permissions.

MKen-li:
Marked as blocked, waiting for bugs

MKen-li:
Marked as verified
Build = dev-209

PR #980, opened 2018-11-20
validation page changes for virtual device adm

AnkRaiza:
#962 
image


src/components/Title.tsx
```diff
@@ -116,15 +120,22 @@ export class Title extends React.Component<TitleProps, any> {
                         <div>
                             <LocaleDropdown disabled={this.props.isYamlEditor} locale={selectedSourceLocale} handleLocaleChange={this.handleLocaleChange} />
                             {
-                                (this.props && this.props.token &&
+                                (this.props && virtualDevices && virtualDevices.length &&
                                     (
                                         <div>
-                                            <label>{this.props.token || "Validation Token"}</label>
+                                            {
+                                                process.env.NODE_ENV === "production" ?
```
jperata: We should remove the toggle if this is going to be deployed-

AnkRaiza: not for now

Issue #926, opened 2018-10-22
The errors messages shouldn't persist after changing between login form and register form


MKen-li:
**Preconditions**
None.

**Steps To Reproduce**
1. Go to Login Page
2. Make the system displayed a validation message like "missing information, wrong password, etc" on login form or register form
3. Change between forms

**Expected behavior**
Once the user changes the form the validation message is deleted.

**Actual behavior**
The message stays on the page

**Screenshots**
https://media.giphy.com/media/ncoLV0zBpS8iUM3exH/giphy.gif

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-162

**Additional context**
None.

MKen-li:
Marked as verified
Build = dev-200

PR #985, opened 2018-11-21
BUG-926: The errors messages shouldn't persist after changing between…

jricaldi:
Relates to #926


src/components/AuthForm/AuthForm.tsx
```diff
@@ -252,11 +252,13 @@ export class NormalLoginForm extends React.Component<NormalLoginFormProps, Norma
...
+        this.state.error = "";
         this.setState(this.state);
     }
 
     onCancelClick() {
         this.state.confirmPassword = "";
+        this.state.error = "";
         this.state.isConfirmPassword = false;
         this.setState(this.state);
     }
```
ecruzado: probably you can create a method for that, function are almost the same

Issue #1011, opened 2018-11-27
When chaining the type of source the virtual device drop-down doesn't have a default value


MKen-li:
**Preconditions**
Create one or more virtual devices of each type

**Steps To Reproduce**
1. Create an Alexa source
2. Go to Validation Page
3. Change the type to Google

**Expected behavior**
The drop-down shows the latest Google token created

**Actual behavior**
The drop-down is blank

**Screenshots**
![blank dropdown](https://user-images.githubusercontent.com/39920441/49111791-20bef300-f25f-11e8-8221-9ed9c7b14935.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-203

**Additional context**
The same behavior can be replicated the other way around.

MKen-li:
Marked as verified
Build = dev-207

Issue #1010, opened 2018-11-27
The position of the Name column changes when you select Alexa 


MKen-li:
**Preconditions**
Have at least one virtual device for Google and other for Alexa

**Steps To Reproduce**
1. Go to Virtual device manager
2. Use the filter and change between Alexa and all or Alexa and Google

**Expected behavior**
The positions of the columns remain the same

**Actual behavior**
The name column changes positions

**Screenshots**
None.

**Environment:**
 - OS: Win10
 - Browser Chrome
 - Version dev-206

**Additional context**
none.

MKen-li:
Closing ticket, working as expected

Issue #1009, opened 2018-11-27
Add a message when the user click the "Create Virtual Device" and has reached the maximum number of virtual devices


MKen-li:
**Preconditions**
have the maximum amount of virtual device

**Steps To Reproduce**
1. Go to Validation page
2. Select the type that has none virtual devices added
3. Click the "Create Virtual Device"

**Expected behavior**
The system shows the next message in a snack bar:
You’ve reached the maximum amount of virtual devices allowed for your plan. Need some more? [Let’s talk]. 

mailto:sales@bespoken.io
Subject: Additional Virtual Devices. 

**Actual behavior**
The permissions popup is displayed and overwrites the latest token created

**Screenshots**
none.

**Environment:**
 - OS: Win 10 
 - Browser Chrome
 - Version dev-206

**Additional context**
The Lets talk is a link to open the email manager.

MKen-li:
Missing subject in email,
moving back to dev

MKen-li:
Mark as verified
Build = dev-208

Issue #1008, opened 2018-11-27
The virtual device name in the Validation passes the drop-down area


MKen-li:
**Preconditions**
Create a virtual device that has 25 characters (max number for a virtual device)

**Steps To Reproduce**
1. Go to Validation Page
2. Expand the virtual device drop down

**Expected behavior**
The name displays 17 chars with 3 dots

**Actual behavior**
The name passes the drop-down area

**Screenshots**
![long name validation page](https://user-images.githubusercontent.com/39920441/49110144-a68c6f80-f25a-11e8-9d28-114108df2c94.png)

**Environment:**
 - OS: Win10
 - Browser Chrome
 - Version dev-206

**Additional context**

MKen-li:
The 3 dots appear only when the drop-down is open but the name still touches the drop-down arrow when is collapse.

MKen-li:
Marked as verified
Build = dev-208

PR #1016, opened 2018-11-29
fix some bugs

AnkRaiza:
Relates #999 #1007 #1008 #1009 #1010 #1011


src/components/GenericInlineEdit.tsx
```diff
@@ -23,15 +23,24 @@ export default class GenericInlineEdit extends React.Component<any, any> {
         }));
     }
 
+    validateString = (value: string) => {
+        return value.trim() && value.trim().length <= 25;
```
jricaldi: `25` should be a constant?

Issue #1019, opened 2018-11-29
Sub-3 (Ref-995) Implement timer 30 days by default


jricaldi:
Relates to #995

MKen-li:
no QA needed

PR #1024, opened 2018-12-04
SUB-1019: Implement timer 30 days by default

jricaldi:
Relates to #1019 

The `TRIAL_TIME` variable can be inserted in the last subtask.

last day:
image

one day:
image

more than 1 day:
image

user without trail standard plan:
image


jperata:
Add screenshots of the 4 possible states (non-trial user, trial user with more than 1 day, trial user with 1 day, trial user on last date).

Issue #1018, opened 2018-11-29
"My account" page title is missing


dmarvp:
**Preconditions**
Be a user of the Dashboard.

**Steps To Reproduce**
1. Go to My account page.

**Expected behavior**
The page should have a title saying "My Account"

**Actual behavior**
No title.

**Screenshots**
Current screen:
<img width="1518" alt="image" src="https://user-images.githubusercontent.com/6411740/49234230-abb9fd80-f3c5-11e8-9e2e-bef5b99c9fc7.png">

Mockup:
<img width="1061" alt="image" src="https://user-images.githubusercontent.com/6411740/49234256-b70d2900-f3c5-11e8-88f5-386e88cd3179.png">

Virtual Device Manager example:
<img width="1516" alt="image" src="https://user-images.githubusercontent.com/6411740/49234313-cf7d4380-f3c5-11e8-8306-c692536f7d01.png">

**Additional context**
Also, make sure that the title refreshes correctly when changing between "My Account", "Virtual Device Manager" and other pages.

MKen-li:
Marked as verified
Build = dev-213

Issue #1017, opened 2018-11-29
My account page not counting the virtual devices correctly


dmarvp:
**Preconditions**
Have more than one virtual device created in the virtual device manager.

**Steps To Reproduce**
1. Go to My account.
2. Look at the count of virtual devices label

**Expected behavior**
The number should correspond to the number of virtual devices created.

**Actual behavior**
If one virtual device has been created using the old way, it will say 1. That number does not change.

**Additional context**
We should count the number of virtual devices under the "virtualDevices" property of the user's object. If we can't find anything there, then we should look for the silentEchoToken to just count 1.

MKen-li:
Marked as verified
Build = dev-213

PR #1027, opened 2018-12-04
fix title and virtual devices number

AnkRaiza:
Relates #1017 #1018


src/pages/MyAccountPage.tsx
```diff
@@ -90,6 +95,7 @@ export class MyAccountPage extends React.Component<MyAccountPageProps, MyAccount
     }
 
     async componentDidMount() {
+        this.props.setPageSubtitle("My account");
```
dmarvp: "My Account"

src/services/subscription-plan.ts
```diff
@@ -38,13 +38,13 @@ export class Payment {
...
-            return userDetail.silentEchoToken ? 1 : 0;
-        }
-
-        userDetail = await auth.currentUserDetails();
-        if (userDetail) {
-            return userDetail.silentEchoToken ? 1 : 0;
+        let details = userDetail;
+        if (!details) details = await auth.currentUserDetails();
+        if (details) {
+            const virtualDevicesLength = process.env.NODE_ENV === "production"
```
dmarvp: I think we don't need the "env" check here. Either they have the virtualDevices property or they have the silentEchoToken :)

AnkRaiza: they could have both and this check is because so far we have the virtual devices on dev only

dmarvp: but if they have both then the ones that matter are the ones from the virtualDevices property right?

Issue #1059, opened 2018-12-13
[Hot fix] The user can only monitor 1 source


MKen-li:
**Preconditions**
None

**Steps To Reproduce**
1. Go to My account (notice that the monitor apps is set to up to 3)
2. Try to enable the 2nd monitoring

**Expected behavior**
The user can enable the monitoring

**Actual behavior**
Payment pop up appears

**Screenshots**
None.

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version prod-1.0.57

**Additional context**
The issue is not reproducible on Dev environment

MKen-li:
Issues verified on dev. waiting for prod release to close the issue

MKen-li:
Verified on staging-47, closing ticket

PR #1062, opened 2018-12-14
BUG-1059: [Hot fix] The user can only monitor 1 source

jricaldi:
Relates to #1059


src/services/subscription-plan.ts
```diff
@@ -173,18 +173,14 @@ export class Payment {
...
-            if (element.source && element.source.validation_enabled) {
-                count++;
-            }
-            if (element && element.validation_enabled) {
-                count++;
-            }
-        }
-        return count;
-    }
+    public getMonitoredSkillNumber = (sources: any[] = []) =>
```
jperata: nice refactor!

Issue #995, opened 2018-11-25
Dashboard 30 day trial


dmarvp:
### Summary

As a new user of the Dashboard
I want to have a 30 day trial period
so that I can try the different functionalities it has

### Acceptance Criteria

1. GIVEN: I'm a new user.
WHEN: I finish registering or login in with Amazon/Github for the first time.
THEN:
- I'm immediately assigned the "Standard" plan
- I see the Dashboard landing page.
- On the top right corner I see the following:
    - Welcome, "user display name"! (if it's not registered yet, use the user's email)
    - Current plan: Standard
    - Only "x" days left. Subscribe now! link. The X should update with the remaining days. The initial number of days should be configurable in an env variable or from the database (default 30).

2. GIVEN: I'm on the dashboard landing page.
WHEN: I click on the "Subscribe now" link.
THEN:  The payment plans modal appears. Note: This time it should allow selecting any plan in the list.

3. GIVEN: I'm a user of the Dashboard.
WHEN: I try to log in after the 30 days of trial.
THEN: I see a screen with the following:
- Hi there! Seems like your trial period has ended...
- Subscribe to one of our paid plans to continue testing your voice apps
- "Subscribe" button
- Do you need more time to evaluate our tools? [Let's get in touch!](mailto:contact@bespoken.io)

4. GIVEN: I'm on the trial period ended screen.
WHEN: I click on the Subscribe button.
THEN: The payment plans modal appears. Note: This time it should allow selecting any plan in the list.

5. GIVEN: I'm on the trial period ended screen.
WHEN: I click on the "Let's get in touch!" link.
THEN: An email client is open:
    - To: contact@bespoken.io
    - Subject: Trial Period Extension

### Additional Resources

![image](https://user-images.githubusercontent.com/6411740/48985624-6ea8ef00-f0d7-11e8-8251-160de8f11474.png)

<img width="1121" alt="image" src="https://user-images.githubusercontent.com/6411740/48985792-e2e49200-f0d9-11e8-89d4-5f477183ca54.png">

jkelvie:
"Seems like your trial period has ended" -> "Your trial period has ended"
"Let's get in touch" -> "Get in touch"
I don't like the message no probllama - I think just showing a timer that is out of time would be better.

MKen-li:
Marked as blocked waiting for:
https://github.com/bespoken/dashboard/issues/1039
https://github.com/bespoken/dashboard/issues/1040
[...]

PR #1064, opened 2018-12-18
[no-ticket]: show trial time from begining(case 0 days)

jricaldi:
Relates to #995,

Depends to https://github.com/bespoken/source-api/pull/155


jperata:
In cases like this when a PR depends of other PR in another project, indicate so by adding the reference to that PR.

PR #1072, opened 2019-01-03
fix false positives

AnkRaiza:


src/pages/validation/ValidationPage.tsx
```diff
@@ -191,6 +191,30 @@ configuration:
         });
     }
 
+    setupYamlChannel = (sourceId: string, timestamp: number) => {
+        if (!process.env.PUSHER_APP_KEY) return;
+        const self = this;
+        const channel = sourceId + timestamp;
+        self.setState(prevState => ({
+            ...prevState,
+            channels: [...self.state.channels, channel]
```
jricaldi: prevState has channels, so try with this. `channels: [...prevState.channels, channel]`

src/pages/validation/ValidationPage.tsx
```diff
@@ -191,6 +191,30 @@ configuration:
         });
     }
 
+    setupYamlChannel = (sourceId: string, timestamp: number) => {
+        if (!process.env.PUSHER_APP_KEY) return;
+        const self = this;
+        const channel = sourceId + timestamp;
+        self.setState(prevState => ({
```
jricaldi: It is not necessary `self` here. just use it in line `209`

PR #1078, opened 2019-01-07
Dropdown on home page modified

daisyzarzo:
relates to 
https://github.com/bespoken/dashboard/issues/610

Issue #610, opened 2018-03-06
Dropdown on home page should redirect to the skill initial page


jperata:
### Expected Behavior
When clicking on a name on the dropdown, it opens the source initial page

### Actual Behavior
It marks the card were the app is but it doesn't 

### Reproduction Steps
Attaching a gif with the reproduction
![dropdownlistissues](https://user-images.githubusercontent.com/8781604/37049013-99c82936-213d-11e8-83ca-b5b55921add9.gif)

MKen-li:
Returning to dev,
The Return home button that appears when the URL is incorrect returns an extra / at the end making this new functionality not to work

PR #1082, opened 2019-01-15
Dropdown on home page redirect to the skill validation page

daisyzarzo:
relates to 
https://github.com/bespoken/dashboard/issues/610
https://github.com/bespoken/dashboard/pull/1078


jperata:
This is the second time this bug have appeared,  it needs a test to ensure it doesn't appear again

src/frames/Dashboard.tsx
```diff
@@ -135,144 +135,148 @@ class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
-      let currentPath = this.props.location.pathname;
-      let newPath = currentPath.replace(this.props.currentSource.id, sourceDropdownableAdapter.source.id);
-      this.props.goTo(newPath);
+    handleSelectedSource = (sourceDropdownableAdapter: SourceDropdownableAdapter, goToPath?: string) => {
+        this.props.setSource(sourceDropdownableAdapter.source);
+        let currentPath = this.props.location.pathname;
+        if (goToPath) {
+            this.props.goTo(goToPath);
+        } else {
+            let newPath = currentPath.replace(this.props.currentSource.id, sourceDropdownableAdapter.source.id);
```
jricaldi: change this to `/skills/${source.value}/validation` as well in case goToPath is empty

jricaldi: after a call, ignore this comment. in some cases, we need to get the current location.

PR #1107, opened 2019-01-24
Table fix

jricaldi:


src/containers/Root.tsx
```diff
@@ -127,6 +126,9 @@ class Root extends React.Component<any, any> {
   renderProvider = () => (
     <Provider store={this.props.store}>
       <Router history={this.props.history} onUpdate={this.onUpdate}>
+        <Route path="/components">
```
jperata: having null or undefined components in this section was causing route issues when deploying to production, is this solved with the webpack changes?

jricaldi: no, you can get null or undefined `child` into a Route, we wont get issues, I tested it.

Issue #1109, opened 2019-01-24
New user's status field


dmarvp:
**Summary**
We need to update a new status property for users in Dashboard, and update current ones.

**Acceptance Criteria**
1. WHEN: A new user registers to the Dashboard.
THEN: 
- His status is set to __trial__
- A trialEndDate property is created (instead of calculating it every time as we do now)

2. WHEN: A user subscribes to a plan
THEN: 
- His status is set to __enabled__
- The user's virtual devices are updated to the enabled status (call the virtual device endpoint for doing that)

__ALSO__: Using the current logic we have, we need to run a script to update current users to the enabled and trial status (with trialEndDate).

**Additional context**
Unsubscribed and disabled status are being handled in other stories.
Trial date handling should be updated to use the trialEndDate property. This will give us flexibility to increase a users trial period if we so desire.

MKen-li:
Marked as verified
Build = dev-266

MKen-li:
Marked as verified
Build = dev-266

PR #1138, opened 2019-02-05
New user's status field

daisyzarzo:
relates to 
https://github.com/bespoken/dashboard/issues/1109


src/components/BillCard/Payment.tsx
```diff
@@ -47,6 +48,9 @@ class SplitFormC extends React.Component<any & SplitFormCProps, any> {
 
     setStripe = async (tokenId: any) => {
         const { email, userId } = this.props.user;
+        console.log("this.props.userDetail");
```
jperata: remove

src/components/BillCard/Payment.tsx
```diff
@@ -47,6 +48,9 @@ class SplitFormC extends React.Component<any & SplitFormCProps, any> {
...
+        console.log(this.props.userDetail);
```
jperata: remove

src/services/auth.ts
```diff
@@ -189,84 +193,88 @@ namespace auth {
...
 
     export const loginWithCustomToken = async (token: string, auth: remoteservice.auth.Auth = remoteservice.defaultService().auth(), localStorage: LocalStorage = new BrowserStorage()): Promise<User> => {
         const user: remoteservice.user.User = await auth.signInWithCustomToken(token);
         const userDetails = await currentUserDetails();
         const isNewUser = !userDetails.originalCreatedAt && !isLegacyUserWithBenefics(userDetails);
+
         if (isNewUser) {
-          await updateCurrentUser({ planId: planByDefaultForNewUser });
+            const props = { planId: planByDefaultForNewUser, status: userStatus.Trial };
+            await createUserDetails(user.uid, props);
```
jperata: just use, source.createUserDetail(user.uid, props)

src/services/auth.ts
```diff
@@ -309,22 +317,26 @@ namespace auth {
...
             const userWithExtraInfo = await source.getUserWithExtraInfo(currentUser.uid);
-            return new UserDetails(silentEchoToken, smAPIAccessToken, vendorID, stripeSubscribedPlanId,
-                stripeCustomerObjId, stripeSubscribedPlanName, planId, name, company, virtualDevices, userWithExtraInfo);
+            return new UserDetails(silentEchoToken, smAPIAccessToken, vendorID, status, stripeSubscribedPlanId,
+                stripeCustomerObjId, stripeSubscribedPlanName, planId, name, company, virtualDevices, userWithExtraInfo, trialEndDate);
         }
         return emptyUser;
     };
 
+    export const createUserDetails = (uid: string, props: Object): Promise<any> => {
```
jperata: remove this function, it doesn't seem to be used outside auth. And you just can call the source

src/services/auth.ts
daisyzarzo: I did it so the test was easy to mock

src/services/auth.ts
daisyzarzo: I did it so the test was easy to mock

PR #1145, opened 2019-02-06
disable continue button on enterprise and restrict by platform

AnkRaiza:


jperata:
expand the tests in virtualdevicepage.test with this scenario

PR #1153, opened 2019-02-08
Login with github fix and falsy validation for plansData

AnkRaiza:


src/index.tsx
```diff
@@ -70,7 +70,8 @@ Firebase.auth().onAuthStateChanged(function (user: Firebase.User) {
     const newLocation = { ...location, ...{ pathname: "/" } }; // Doing this will pass along any query parameters that may exist.
     // If there is a user, set it
     if (user) {
-        if (!user.emailVerified) {
+        const isGithubUser = user.providerData && user.providerData[0] && user.providerData[0].providerId === "github.com";
```
jperata: there is code in line 81 that does the same, but for all providers that don't include a password, it seems like extracting that validation here resolves the issue, because it seems this might cause the issue on google or amazon logins

Issue #1095, opened 2019-01-18
Renew subscription page


dmarvp:
### Summary

As a user that has canceled his subscription
I want to be able to renew my subscription
When I try to login into the Dashboard

### Acceptance Criteria

1. GIVEN: I'm a user that has canceled his subscription
WHEN: I try to login into the Dashboard
THEN: I see the following:
- The same layout as the trial ended page
- Text: 
> ### Hi there!
> We are glad to see you again
> [ graphic ]
> Subscribe to one of our paid plans to continue testing your voice apps
> [Subscribe](url)

2. GIVEN: I'm on the page 
WHEN: I click on the "Subscribe" button
THEN: I see the payment modal.

3. GIVEN: I subscribe to one of the paid plans
WHEN: I close the modal
THEN: The page refreshes and I can see
    - My previously created sources
    - Monitoring enabled for sources that had been monitored before
    - Virtual devices I had before should be enabled again (call the virtual device endpoint to do that)

### Additional Resources

* TBD

### NOTES
For this to work, when unsubscribing from a plan, we need to have saved the data of previously monitored sources and which platform they belonged to.

Related to https://github.com/bespoken/dashboard/issues/965

MKen-li:
Marked as blocked waiting for https://github.com/bespoken/technical-debt/issues/8

MKen-li:
Marked as verified
Build  = dev-291

PR #1154, opened 2019-02-08
FEAT-1095: Renew subscription page

jricaldi:
Relates to #1095 

image


jperata:
the tabs should not be present in this screen

jperata:
code is good but missing some tests

Issue #1157, opened 2019-02-08
The virtual device tab shouldn't be hidden when no sources are created


MKen-li:
**Preconditions**
None.

**Steps To Reproduce**
1. Log in to the system with a user that has 0 sources created

**Expected behavior**
The virtual device tab is present and the user can create a virtual device without having any sources created

**Actual behavior**
Virtual devices are only available after creating a source

**Screenshots**
![virtual device tab](https://user-images.githubusercontent.com/39920441/52507299-1efa9f80-2bbf-11e9-8381-c6149b7a0e44.png)

**Environment:**
 - OS: Win 10
 - Browser Chrome
 - Version dev-267

**Additional context**
None.

MKen-li:
Marked as verified
Build = dev-274

PR #1171, opened 2019-02-12
virtual device tab allowed when there is no source or skill

daisyzarzo:
Fixes #

relates to  https://github.com/bespoken/dashboard/issues/1157

image


jperata:
small extra assertion on the headers.test.ts for the virtual device tab component

PR #1174, opened 2019-02-13
add userDetails to store state

AnkRaiza:


src/pages/BillPage.tsx
```diff
@@ -0,0 +1,117 @@
+import * as moment from "moment";
```
jperata: It seems like as part of your merge you re-added this page. 
This page was removed from the code and the routes (this is probably also taking down your coverage)

PR #1223, opened 2019-03-01
show stripe error based on server response

AnkRaiza:


src/services/stripe.ts
```diff
@@ -56,15 +56,15 @@ export async function registerSubscription(user: User, tokenId: any, planToSubsc
             },
             body: query.json()
         });
-
         if (result.status === 200) {
             return "operation success";
         }
-        if (result.status === 401) {
-            return "The request has not been applied because it lacks valid authentication ";
+        const data = await result.json();
```
jricaldi: Is this not necessary for line 59?

AnkRaiza: for the successful operation we only need te status code, we dont use the body or any data from it

PR #1233, opened 2019-03-01
Switch update approach

AnkRaiza:
switch to a for loop since the promise all overlaps and fails to update virtual devices correctly


src/services/virtualDevice.ts
```diff
@@ -1,30 +1,15 @@
 import source from "../services/source";
 
-export const enableUserVirtualDevices = async (userId: string, virtualDevices: any) => {
+export const updateUserVirtualDevicesStatus = async (userId: string, virtualDevices: any, status: string) => {
     try {
-        const virtualDevicePromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDeviceStatus(vd.token, "enabled"));
-        await Promise.all(virtualDevicePromises);
-        const virtualDeviceFirebasePromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDevice(vd.token, userId, undefined, "enabled"));
-        await Promise.all(virtualDeviceFirebasePromises);
+        const virtualDevicePromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDeviceStatus(vd.token, status));
```
jperata: since you are waiting for all the promises, you don't need to await here

src/services/virtualDevice.ts
```diff
@@ -1,30 +1,15 @@
...
+        const virtualDeviceFirebasePromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDevice(vd.token, userId, undefined, status));
+        await Promise.all(virtualDevicePromises.concat(virtualDeviceFirebasePromises));
         return Promise.resolve();
```
jperata: async functions , can just return, or in this case since there is no return value, you can just finish (it will still return a promise that resolve in undefined)

src/services/virtualDevice.ts
```diff
@@ -1,30 +1,15 @@
...
-        const virtualDeviceDynamoPromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDeviceStatus(vd.token, "suspended"));
-        await Promise.all(virtualDeviceDynamoPromises);
-        const virtualDeviceFirebasePromises = virtualDevices.map(async (vd: any) => await source.updateVirtualDevice(vd.token, userId, undefined, "suspended"));
-        await Promise.all(virtualDeviceFirebasePromises);
-        return Promise.resolve();
-    } catch (error) {
-        const errorMessage = error.response && error.response.data && error.response.data.error + " ";
-        const token = error.config && error.config.url + " ";
-        return Promise.reject("There was a problem trying to * suspended the virtual device." + errorMessage + token);
+        return Promise.reject("There was a problem trying to update the virtual device." + errorMessage + token);
```
jperata: proper throw instead of rejecting since it's using async/await

PR #1274, opened 2019-03-25
change on single plan for header

AnkRaiza:


src/components/Header.tsx
```diff
@@ -98,6 +98,9 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
     }
 
     componentWillReceiveProps(nextProps: HeaderProps, context: any) {
+        const prevPlatform = this.props.userDetails && this.props.userDetails.benefits && this.props.userDetails.benefits.platforms && this.props.userDetails.benefits.platforms.length && this.props.userDetails.benefits.platforms[0];
```
jperata: create method 
extractPlatformFromUserDetails = (userDetails) => {
   return userDetails && userDetails.benefits && userDetails.benefits.platforms && userDetails.benefits.platforms.length && userDetails.benefits.platforms[0]; 
}

and do
         const prevPlatform = extractPlatformFromUserDetails(  this.props.userDetails )
         const nextPlatform = extractPlatformFromUserDetails(  nextProps.userDetails )

